PR #7, opened 2023-08-15
Fix build

jack-berg:


buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
```diff
@@ -105,5 +105,6 @@ tasks {
 configurations.configureEach {
   resolutionStrategy {
     failOnVersionConflict()
+    preferProjectModules()
```
jack-berg, line 108: The build is [failing](https://github.com/open-telemetry/semantic-conventions-java/actions/runs/5869187161/job/15913649095) without this 🤔

trask (approved):

PR #9, opened 2023-08-15
Update to semantic conventions

jack-berg:
- Update codegen to point at `semantic-conventions` instead of `opentelemetry-specification`
- Update to v1.21.0 of `semantic-conventions`
- Continue using strategy of retaining deleted attributes with `@Deprecated` annotation

### DIFF
--- a/buildscripts/templates/SemanticAttributes.java.j2
+++ b/buildscripts/templates/SemanticAttributes.java.j2
@@ -69,6 +69,8 @@ public final class {{class}} {
     {%- if attribute.note %}
   *
   * <p>Notes:
+        {# NOTE: replace("> ", "") removes the following problematic characters which produce mangled javadoc:  #}
+        {# https://github.com/open-telemetry/semantic-conventions/blob/c83a10a9c33c18a769835e959200d0e24dc708fe/model/resource/k8s.yaml#L34-L38  #}
         <ul> {{attribute.note | replace("> ", "") | render_markdown(code="{{@code {0}}}", paragraph="<li>{0}</li>", list="{0}")}} </ul>
 
     {%- endif %}
--- a/buildscripts/templates/SemanticAttributes.java.j2
+++ b/buildscripts/templates/SemanticAttributes.java.j2
@@ -94,16 +94,6 @@ public final class {{class}} {
 
     {%- endfor %}
 
-    {%- if class_name == "NetTransportValues" %}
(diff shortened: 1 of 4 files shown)


src/main/java/io/opentelemetry/semconv/resource/attributes/ResourceAttributes.java
```diff
@@ -128,7 +128,7 @@ public final class ResourceAttributes {
    *       href="https://docs.microsoft.com/en-us/rest/api/resources/resources/get-by-id">Fully
    *       Qualified Resource ID</a> of the invoked function, <em>not</em> the function app, having
    *       the form {@code
-   *       /subscriptions/<SUBSCIPTION_GUID>/resourceGroups/<RG>/providers/Microsoft.Web/sites/<FUNCAPP>/functions/<FUNC>}.
+   *       /subscriptions/<SUBSCIPTION_GUID/resourceGroups/<RG/providers/Microsoft.Web/sites/<FUNCAPP/functions/<FUNC}.
```
breedx-splk, line 131: This looks like it got a little mangled along the way.

breedx-splk (approved):
Cool stuff, I commented on one oddity. 🤷🏻

trask (approved):

src/main/java/io/opentelemetry/semconv/resource/attributes/ResourceAttributes.java
jack-berg: Nice catch

buildscripts/templates/SemanticAttributes.java.j2
```diff
@@ -69,7 +69,8 @@ public final class {{class}} {
     {%- if attribute.note %}
   *
   * <p>Notes:
-    <ul> {{attribute.note | render_markdown(code="{{@code {0}}}", paragraph="<li>{0}</li>", list="{0}")}} </ul>
+        <ul> {{attribute.note | replace("> ", "") | render_markdown(code="{{@code {0}}}", paragraph="<li>{0}</li>", list="{0}")}} </ul>
```
jack-berg, line 72: FYI, the `replace("> ", "")` is to remove [this](https://github.com/open-telemetry/semantic-conventions/blob/c83a10a9c33c18a769835e959200d0e24dc708fe/model/resource/k8s.yaml#L34-L38), which causes the generator to produce mangled javadoc which fails the build.

mateuszrzeszutek (approved):

PR #11, opened 2023-08-16
Add publishing conventions and release workflow

### LINKED ISSUE #3: Setup release workflow
Build release workflow to publish artifacts to [sonotype](https://oss.sonatype.org/index.html#stagingRepositories).

Should model after `opentelemetry-java`, but consider streamlining if any of the complexity of that setup can be avoided.

jack-berg:
Resolves #3.

The release requirements for this repository are different than `opentelemetry-java`, `opentelemetry-java-instrumentation`, `opentelemetry-java-contrib` and `opentelemetry-proto-java`:

- Unlike core, instrumentation, and contrib, the release version is not autoincrementing (not for major, minor, or patch). If we release 1.21.0, we can't update the _next version_ to be 1.22.0 because that release of `opentelemetry/semantic-conventions` will almost always not exist yet.
- Instead, I think we will have to manually increment the corresponding version of `opentelemetry/semantic-conventions`. After manually updating, we can make any required manual tweaks to the generated code (i.e. adding deprecations) and eventually run the release workflow.
- The release workflow comes in two stages:
  * First, run `./.github/workflows/prepare-release-branch.yml`, which:
    * Validate on main branch, and that CHANGELOG.md has unreleased section
    * Create the release branch (i.e. `release/v1.21.0`), first verifying that it does not already exist (this might be an easy mistake to make if you try to run the workflow without manually incrementing the version of `opentelemetry/semantic-conventions`
    * Create a PR against the release branch to change the snapshot boolean in `build.gradle.kts` from true to false, and to update the changelog with the release date.
    * Create a PR against main to update the chnagelog with the release date.
    * Merge both PRs.
  * Next, run `./.github/workflows/release.yml` against the release branch, which:
    * runs `./gradlew assemble publishToSonotype closeAndReleaseSonatypeStagingRepository`
    * Publishes release notes, including the updates from the changelog and the list of contributors.
- Not that there is no auto-updating the `opentelemetry/semantic-conventions` version number in the release. This means that before the release, we'll publish to the same snapshot tags as after. This is a bit weird and we need to think this through. Maybe we don't publish snapshots in this repo?

Reminder for myself to add required secrets to github: `SONATYPE_USER`, `SONATYPE_KEY`, `GPG_PRIVATE_KEY`, `GPG_PASSWORD`

Should also document the process, but probably in a followup after the first release.


breedx-splk (approved):
Looks good to me. Anything specific to look at that's not copypasta?

.github/scripts/generate-release-contributors.sh
```diff
@@ -0,0 +1,90 @@
+#!/bin/bash -e
```
jack-berg, line 1: This file is copy / paste from `opentelemetry-java`.

.github/scripts/get-version.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash -e
+
+grep -Eo "var semanticConventionsVersion = \".*\"" build.gradle.kts | grep -Eo "[0-9]+.[0-9]+.[0-9]+"
```
jack-berg, line 3: Modified to find the version from `build.gradle.kts`, and needs to pick out the correct version since there are many matches for the `[0-9]+.[0-9]+.[0-9]+` pattern in that file.

.github/scripts/use-cla-approved-github-bot.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash -e
```
jack-berg, line 1: This file is copy / paste from opentelemetry-java.

.github/workflows/prepare-release-branch.yml
```diff
@@ -0,0 +1,115 @@
...
+      - name: Create release branch
+        run: |
+          version=$(.github/scripts/get-version.sh)
+          if [[ $version =~ ^([0-9]+)\.([0-9]+)\.0$ ]]; then
+            release_branch_name="release/v${version}"
+            git ls-remote --exit-code --heads origin refs/heads/$release_branch_name
+            if [ "$?" == "0" ] ; then 
+              echo "release branch $release_branch_name already exists"
+              exit 1 
+            fi
```
jack-berg, line 39: In otel java, release_branch_name is set to `release_branch_name=release_branch_name=$(echo $version | sed -E 's/([0-9]+)\.([0-9]+)\.0/release\/v\1.\2.x/')`, which is more complicated than needed for this repo.

In this repo, we confirm the release branch doesn't already exist, which is needed to prevent duplicate releases since the version of `open-telemetry/semantic-convention` is manually updated:
```            
git ls-remote --exit-code --heads origin refs/heads/$release_branch_name
if [ "$?" == "0" ] ; then 
  echo "release branch $release_branch_name already exists"
  exit 1 
fi
```

.github/workflows/prepare-release-branch.yml
```diff
@@ -0,0 +1,115 @@
...
+          branch="opentelemetrybot/prepare-release-${VERSION}"
+
+          git checkout -b $branch
+          git commit -a -m "$message"
+          git push --set-upstream origin $branch
+          gh pr create --title "[$RELEASE_BRANCH_NAME] $message" \
+                       --body "$message." \
+                       --base $RELEASE_BRANCH_NAME
+
+  create-pull-request-against-main:
```
jack-berg, line 76: This is the same as in `opentelemetry-java`, except we skip the "Update version" step since the version of `open-telemetry/semantic-convention` is manually updated.

.github/workflows/release.yml
```diff
@@ -0,0 +1,114 @@
...
+        # back to the release branch
+      - uses: actions/checkout@v3
+        with:
+          # tags are needed for the generate-release-contributors.sh script
+          fetch-depth: 0
+
+      - name: Generate release notes
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
```
jack-berg, line 82: This skips the special patch release logic which adds a preamble.

.github/workflows/release.yml
```diff
@@ -0,0 +1,114 @@
...
+
+          # conditional block not indented because of the heredoc
+          cat >> /tmp/release-notes.txt << EOF
+
+          ### 🙇 Thank you
+          This release was possible thanks to the following contributors who shared their brilliant ideas and awesome pull requests:
+
+          EOF
+
+          .github/scripts/generate-release-contributors.sh v$PRIOR_VERSION >> /tmp/release-notes.txt
```
jack-berg, line 101: We add contributors on all releases, not just minor releases.

build.gradle.kts
```diff
@@ -1,13 +1,32 @@
...
+val snapshot = true
+// end
+
+// The release version of https://github.com/open-telemetry/semantic-conventions used to generate classes
+var semanticConventionsVersion = "1.21.0"
+
+// Compute the artifact version, which includes the "-alpha" suffix and includes "-SNAPSHOT" suffix if not releasing
+// Release example: version=1.21.0-alpha
+// Snapshot example: version=1.21.0-alpha-SNAPSHOT
+var releaseVersion = semanticConventionsVersion + "-alpha"
```
jack-berg, line 24: The `-alpha` suffix is hardcoded until we're ready to publish a stable release.

jack-berg:
@breedx-splk I've annotated the PR with comments which describe any differences from the `opentelemetry-java` release process.

Issue #5, opened 2023-08-11
Develop strategy for generating semantic conventions in different signals, different domains


jack-berg:
We currently put all semantic convention resource attributes in `io.opentelemetry.semconv.resource.attributes.ResourceAttributes`, and all others in  `io.opentelemetry.semconv.trace.attributes.SemanticAttributes`.

There are several shortcomings with this strategy:
- The `*.trace.*` portion of the package is a misnomer, since SemanticAttributes contains attributes for all signals
- There are going to be semantic convention domains at varying levels of stability (i.e. http is looking to go stable in the fall but the others are still experimental), and even within a single domain different attributes may be at different levels of severity
- The names of metrics are part of the semantic conventions and don't appear anywhere in the generated code

jack-berg:
Proposal from 12/07/23 Java SIG and @lmolkova 's prototype #40:

- Stop differentiating between ResourceAttributes and SemanticAttributes. Attributes are attributes regardless of whether they're used on a resource, log, trace, metric, etc. 
- Break out attributes into classes named after the top level namespace. I.e. `UserAgentAttributes` for all attributes nested under `user_agent.*`. The hope is that this strikes a nice balance and avoids cumbersome mega-classes. 
- Break out attributes from anything else generated from conventions. I.e. if we generate constants for metric names (or functions for constructing instruments), put those in separate classes from the attributes for that namespace. For example, for `http.*` attributes / metrics we would have `HttpAttributes` and `HttpMetrics`.

jack-berg:
Fixed in #45.

Issue #2, opened 2023-08-11
Decide artifact name and versioning scheme


jack-berg:
The existing semconv artifact is published under coordinates `io.opentelemetry:opentelemetry-semconv:{version}`, with the latest version being `1.29.0-alpha`. 

One of the primary reasons for creating a new semantic-conventions-java repo is so that it can version in lockstep with `open-telemetry/semantic-conventions`, whose latest release is `v1.21.0`.

If we use the same artifact coordinates and publish version in lockstep with the semantic-conventions repo, we'll have a conflict. I.e. when we publish a version corresponding to the next version of semantic-conventions v1.22.0, we'll have `io.opentelemetry:opentelemetry-semconv:1.22.0-alpha`, which already exists.

We should consider publishing under a new set of artifact coordinates. Perhaps we modify the group to be `io.opentelemetry.semconv`. This would align with other otel java repos and and allow us to avoid the conflict.

There's also the question discussed [here](https://github.com/open-telemetry/community/issues/1638#issuecomment-1672679808). Should we create a versioning strategy to allow us to change the code generation logic independently of a new version of semantic conventions? 

One way to do this would be to have versions of the form `{CODE_GEN_VERSION}_{SEMCONV_VERSION}`. E.g. `1.0.0_1.21.0`, or perhaps more simply `1_1.21.0`.

jack-berg:
Option 1: Single version synchronized with `semantic-conventions` releases, i.e. `v1.21.0`. Encode code gen version in package name `io.opentelemetry.semconv.v1_0.*`.
- Pro: Simple, flexible for 
- Con: If we change code gen logic, we'll need to continue publishing old generated code package forever, since we have no way to indicate breaking change in the artifact version.

Option 2: Multiple versions, each following semconv, i.e. `v1.0.0_1.21.0`
- Con: confusing and unfamiliar to users
- Con: Probably overkill. Since this repo doesn't depend on anything besides `opentelemetry-api` (for `AttributeKey<?>`), we shouldn't ever need to patch for security vulnerabilities. Any additional attributes we'd want to publish would have a corresponding version of `semantic-conventions`, so probably no need for minor.
- Pro: Preserves the ability to make major (breaking), minor, and patch changes to code gen logic if we ever find the need

Option 3: Multiple versions, with simplified code gen version, i.e. `v1.0_1.21.0`
- Con: confusing and unfamiliar to users. Not obvious why two versions and why one doesn't follow semantic versioning.
[...]

PR #13, opened 2023-08-30
Move to io.opentelemetry.semconv package

jack-berg:
Related to #5, #2.

- `SemanticAttributes` now lives in a package which doesn't include `*.trace.*`, which is accurate given that includes attributes used across signals.
- `ResourceAttributes` lives in the same package as `SemanticAttributes` because the additional level to the package hierarchy serves no value.
- ~~The package includes `v1` to give allow us to make breaking changes to the code generation logic as discussed in [comment](https://github.com/open-telemetry/semantic-conventions-java/issues/2#issuecomment-1699523236). We are still publishing an `-alpha` artifact, so we are allowed to make breaking changes to the code generation logic, but I figure its best to keep the package change churn to a minimum, so trying to get us close to the structure we are likely to use when we publish a stable artifact.~~ Discussed in a comment and decided to optimistically omit v1 from the package name. If later we need to make breaking changes after we release a stable version, we can skip to include v2 in the package name.
- The downside of this set of changes is that the package change will be a breaking change vs. the existing `io.opentelemetry:opentelemetry-semconv` artifact. I think that's fine tho and we can publish deprecation warnings that point to the new classes in `io.opentelemetry:opentelemetry-semconv` before deleting.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -93,7 +93,7 @@ val generateSemanticAttributes by tasks.registering(Exec::class) {
     "--rm",
     "-v", "$buildDir/semantic-conventions/model:/source",
     "-v", "$projectDir/buildscripts/templates:/templates",
-    "-v", "$projectDir/src/main/java/io/opentelemetry/semconv/v1/:/output",
+    "-v", "$projectDir/src/main/java/io/opentelemetry/semconv/:/output",
     "otel/semconvgen:$generatorVersion",
     "--only", "span,event,attribute_group,scope",
     "-f", "/source", "code",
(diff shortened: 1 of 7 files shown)


trask:
> * The package includes `v1` to give allow us to make breaking changes to the code generation logic as discussed in [comment](https://github.com/open-telemetry/semantic-conventions-java/issues/2#issuecomment-1699523236).

what do you think of being optimistic and not including the `v1` in the package name, and only adding `v2` in the future if needed?

jack-berg:
>> The package includes v1 to give allow us to make breaking changes to the code generation logic as discussed in https://github.com/open-telemetry/semantic-conventions-java/issues/2#issuecomment-1699523236.

>what do you think of being optimistic and not including the v1 in the package name, and only adding v2 in the future if needed?

I'm fine with that too. Was hoping this PR would spur that conversation 🙂

trask (approved):

mateuszrzeszutek (approved):

PR #14, opened 2023-08-30
Add generate build step to avoid uncomitted changes

jack-berg:


jack-berg:
I pushed a [commit](https://github.com/open-telemetry/semantic-conventions-java/pull/14/commits/e54077fa0bad2a261846c4ce26049afdd3803602) which changes the template and verified that the [build fails](https://github.com/open-telemetry/semantic-conventions-java/actions/runs/6029172560/job/16358074059?pr=14) when the changes to the generated classes are not committed.

Now reverting that change.

mateuszrzeszutek (approved):

PR #17, opened 2023-09-01
Fix release workflow

jack-berg:
I believe this change will fix this failure: https://github.com/open-telemetry/semantic-conventions-java/actions/runs/6050108339/job/16418698284

Also, adding change log entry.

### DIFF
--- a/.github/workflows/prepare-release-branch.yml
+++ b/.github/workflows/prepare-release-branch.yml
@@ -28,14 +28,15 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: Create release branch
+        shell: bash {0}
         run: |
           version=$(.github/scripts/get-version.sh)
-          if [[ $version =~ ^([0-9]+)\.([0-9]+)\.0$ ]]; then
+          if [[ $version =~ ^([0-9]+)\.([0-9]+)\.([0-9]+)$ ]]; then
             release_branch_name="release/v${version}"
-            git ls-remote --exit-code --heads origin refs/heads/$release_branch_name
-            if [ "$?" == "0" ] ; then 
+            release_branch_exists=$(git ls-remote --exit-code --heads origin refs/heads/$release_branch_name)
+            if [[ $release_branch_exists != "" ]] ; then
               echo "release branch $release_branch_name already exists"
-              exit 1 
+              exit 1
             fi
           else
             echo "unexpected version: $version"
--- a/.github/workflows/prepare-release-branch.yml
+++ b/.github/workflows/prepare-release-branch.yml
@@ -28,12 +28,11 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: Create release branch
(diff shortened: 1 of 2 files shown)


mateuszrzeszutek (approved):

.github/workflows/prepare-release-branch.yml
```diff
@@ -28,14 +28,15 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: Create release branch
+        shell: bash {0}
```
trask, line 31: can you add a comment here why this is needed?

.github/workflows/prepare-release-branch.yml
```diff
@@ -28,14 +28,15 @@ jobs:
...
         run: |
           version=$(.github/scripts/get-version.sh)
-          if [[ $version =~ ^([0-9]+)\.([0-9]+)\.0$ ]]; then
+          if [[ $version =~ ^([0-9]+)\.([0-9]+)\.([0-9]+)$ ]]; then
             release_branch_name="release/v${version}"
-            git ls-remote --exit-code --heads origin refs/heads/$release_branch_name
-            if [ "$?" == "0" ] ; then 
+            release_branch_exists=$(git ls-remote --exit-code --heads origin refs/heads/$release_branch_name)
+            if [[ $release_branch_exists != "" ]] ; then
```
trask, line 37: what was the issue here? (so we can fix in other repos also if needed)

mateuszrzeszutek: I don't think we have that `git ls-remote` check in other repos; probably the shell used `set -e` and `git ls-remote` returned non-zero exit code

jack-berg: This is a check that only exists in this repo. The other repos have a process in which the version is automatically incremented which is reflected in the release branch. In this repo, its much easier to accidentally run the release workflow against main without having manually bumped the `semantic-conventions` release version. This check helps avoid that.

.github/workflows/prepare-release-branch.yml
jack-berg: Was doing this to avoid failing the step if a command has a non-zero exit code. Realized I could fix that without this by just removing the `--exit-code` flag from the `git ls-remote` command so it returns without an exist code if it doesn't find the ref.

.github/workflows/prepare-release-branch.yml
jack-berg: > I don't think we have that git ls-remote check in other repos; probably the shell used set -e and git ls-remote returned non-zero exit code

Exactly, but I pushed a commit to avoid the non-zero exit code, as I mentioned [here](https://github.com/open-telemetry/semantic-conventions-java/pull/17#discussion_r1313171543).

Figure its safer to fail on non-zero exit codes if possible.

trask (approved):

PR #24, opened 2023-09-21
Extended jinja template to generate template-type semantic attributes.

AlexanderWert:
With https://github.com/open-telemetry/build-tools/pull/186, template-type attributes (such as `http.request.header.<key>`) can be defined in semantic convention model yaml files. With that, these template attributes can be used in code generation for the semantic conventions.

This PR adds a new class `AttributeKeyTemplate` for this kind of attribute keys and extends the code generation to cover attributes like `http.request.header.<key>`, `db.elasticsearch.path_parts.<key>`, `rpc.grpc.request.metadata.<key>`, etc.

With this change utility classes like  [`CapturedHttpHeadersUtil`](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api-semconv/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/CapturedHttpHeadersUtil.java#L36) in the java-instrumentation repo can be replaced with instances (and corresponding constants) of the more generic `AttributeKeyTemplate` class.

### DIFF
--- a/src/main/java/io/opentelemetry/semconv/AttributeKeyTemplate.java
+++ b/src/main/java/io/opentelemetry/semconv/AttributeKeyTemplate.java
@@ -6,6 +6,7 @@
 package io.opentelemetry.semconv;
 
 import io.opentelemetry.api.common.AttributeKey;
+import io.opentelemetry.api.common.AttributeType;
 import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
@@ -32,36 +33,68 @@ public final class AttributeKeyTemplate<T> {
     this.keyBuilder = keyBuilder;
   }
 
-  static AttributeKeyTemplate<String> stringKeyTemplate(String prefix) {
-    return new AttributeKeyTemplate<String>(prefix, AttributeKey::stringKey);
+  /**
+   * Create an {@link AttributeKeyTemplate} with type {@link AttributeType#STRING} and the given
+   * {@code prefix}.
+   */
+  public static AttributeKeyTemplate<String> stringKeyTemplate(String prefix) {
+    return new AttributeKeyTemplate<>(prefix, AttributeKey::stringKey);
   }
 
-  static AttributeKeyTemplate<List<String>> stringArrayKeyTemplate(String prefix) {
-    return new AttributeKeyTemplate<List<String>>(prefix, AttributeKey::stringArrayKey);
+  /**
+   * Create an {@link AttributeKeyTemplate} with type {@link AttributeType#STRING_ARRAY} and the
+   * given {@code prefix}.
+   */
+  public static AttributeKeyTemplate<List<String>> stringArrayKeyTemplate(String prefix) {
+    return new AttributeKeyTemplate<>(prefix, AttributeKey::stringArrayKey);
   }
 
-  static AttributeKeyTemplate<Boolean> booleanKeyTemplate(String prefix) {
-    return new AttributeKeyTemplate<Boolean>(prefix, AttributeKey::booleanKey);
+  /**
+   * Create an {@link AttributeKeyTemplate} with type {@link AttributeType#BOOLEAN} and the given
+   * {@code prefix}.
+   */
+  public static AttributeKeyTemplate<Boolean> booleanKeyTemplate(String prefix) {
+    return new AttributeKeyTemplate<>(prefix, AttributeKey::booleanKey);
   }
 
-  static AttributeKeyTemplate<List<Boolean>> booleanArrayKeyTemplate(String prefix) {
-    return new AttributeKeyTemplate<List<Boolean>>(prefix, AttributeKey::booleanArrayKey);
+  /**
+   * Create an {@link AttributeKeyTemplate} with type {@link AttributeType#BOOLEAN_ARRAY} and the
+   * given {@code prefix}.
+   */
+  public static AttributeKeyTemplate<List<Boolean>> booleanArrayKeyTemplate(String prefix) {
+    return new AttributeKeyTemplate<>(prefix, AttributeKey::booleanArrayKey);
   }
 
-  static AttributeKeyTemplate<Long> longKeyTemplate(String prefix) {
-    return new AttributeKeyTemplate<Long>(prefix, AttributeKey::longKey);
+  /**
+   * Create an {@link AttributeKeyTemplate} with type {@link AttributeType#LONG} and the given
+   * {@code prefix}.
+   */
+  public static AttributeKeyTemplate<Long> longKeyTemplate(String prefix) {
+    return new AttributeKeyTemplate<>(prefix, AttributeKey::longKey);
   }
 
-  static AttributeKeyTemplate<List<Long>> longArrayKeyTemplate(String prefix) {
-    return new AttributeKeyTemplate<List<Long>>(prefix, AttributeKey::longArrayKey);
+  /**
+   * Create an {@link AttributeKeyTemplate} with type {@link AttributeType#LONG_ARRAY} and the given
+   * {@code prefix}.
+   */
+  public static AttributeKeyTemplate<List<Long>> longArrayKeyTemplate(String prefix) {
+    return new AttributeKeyTemplate<>(prefix, AttributeKey::longArrayKey);
   }
 
-  static AttributeKeyTemplate<Double> doubleKeyTemplate(String prefix) {
-    return new AttributeKeyTemplate<Double>(prefix, AttributeKey::doubleKey);
+  /**
+   * Create an {@link AttributeKeyTemplate} with type {@link AttributeType#DOUBLE} and the given
+   * {@code prefix}.
+   */
+  public static AttributeKeyTemplate<Double> doubleKeyTemplate(String prefix) {
+    return new AttributeKeyTemplate<>(prefix, AttributeKey::doubleKey);
   }
 
-  static AttributeKeyTemplate<List<Double>> doubleArrayKeyTemplate(String prefix) {
-    return new AttributeKeyTemplate<List<Double>>(prefix, AttributeKey::doubleArrayKey);
+  /**
+   * Create an {@link AttributeKeyTemplate} with type {@link AttributeType#DOUBLE_ARRAY} and the
+   * given {@code prefix}.
+   */
+  public static AttributeKeyTemplate<List<Double>> doubleArrayKeyTemplate(String prefix) {
+    return new AttributeKeyTemplate<>(prefix, AttributeKey::doubleArrayKey);
   }
 
   private AttributeKey<T> createAttributeKey(String keyName) {
--- a/src/main/java/io/opentelemetry/semconv/AttributeKeyTemplate.java
+++ b/src/main/java/io/opentelemetry/semconv/AttributeKeyTemplate.java
@@ -14,9 +14,9 @@ import java.util.function.Function;
 
 public final class AttributeKeyTemplate<T> {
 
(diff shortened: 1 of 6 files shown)


AlexanderWert:
@open-telemetry/java-approvers PTAL, I think this would allow to replace some custom helpers in the agent with generated constants for this type of attributes

trask:
cc @breedx-splk @mateuszrzeszutek to review new public API surface if you have a chance

src/main/java/io/opentelemetry/semconv/AttributeKeyTemplate.java
```diff
@@ -0,0 +1,66 @@
...
+  static AttributeKeyTemplate<Double> doubleKeyTemplate(String prefix) {
+    return new AttributeKeyTemplate<Double>(prefix, AttributeKey::doubleKey);
+  }
+
+  static AttributeKeyTemplate<List<Double>> doubleArrayKeyTemplate(String prefix) {
+    return new AttributeKeyTemplate<List<Double>>(prefix, AttributeKey::doubleArrayKey);
+  }
+
+  private AttributeKey<T> createAttributeKey(String keyName) {
+    String key = prefix + "." + keyName.toLowerCase(Locale.ROOT).replace('-', '_');
```
mateuszrzeszutek, line 59: I have mixed feelings about encoding the `.replace('-', '_')` part here; it was recently removed from the HTTP headers semconv. Perhaps doing any sort of normalization should be the responsibility of the caller of this class?

src/main/java/io/opentelemetry/semconv/AttributeKeyTemplate.java
```diff
@@ -0,0 +1,66 @@
...
+import io.opentelemetry.api.common.AttributeKey;
+import java.util.List;
+import java.util.Locale;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.function.Function;
+
+public final class AttributeKeyTemplate<T> {
+
+  private final ConcurrentMap<String, AttributeKey<T>> keysCache = new ConcurrentHashMap<>();
```
mateuszrzeszutek, line 17: WDYT about a single static cache for all the templates?

src/main/java/io/opentelemetry/semconv/AttributeKeyTemplate.java
```diff
@@ -0,0 +1,66 @@
...
+package io.opentelemetry.semconv;
+
+import io.opentelemetry.api.common.AttributeKey;
+import java.util.List;
+import java.util.Locale;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.function.Function;
+
+public final class AttributeKeyTemplate<T> {
```
mateuszrzeszutek, line 15: Can you add some javadocs to the public methods of this class?

src/main/java/io/opentelemetry/semconv/AttributeKeyTemplate.java
breedx-splk: Yeah, and then prefix the map key with the actual prefix?

breedx-splk (approved):
This looks good to me. Just have the one idea around the key cache being static, but otherwise seems like a nice addition.

src/main/java/io/opentelemetry/semconv/AttributeKeyTemplate.java
mateuszrzeszutek: Yeah, exactly that.

jack-berg: What's the advantage of single map? Sounds like more contention and more resizing outside of calls for a particular prefix. I.e. if i call `HTTP_REQUEST_HEADER.getAttributeKey(..)` many times with distinct names but `RPC_GRPC_RESPONSE_METADATA.getAttributeKey(..)` for a small set of keys, the resizing of the shared map triggered by the cardinality of `HTTP_REQUEST_HEADER` could impact `RPC_GRPC_RESPONSE_METADATA`.

I suppose using a single map minimizes the memory footprint of initializing many maps, most of which won't get used for a particular application. But we could solve that by initializing the maps with a small initial capacity.

src/main/java/io/opentelemetry/semconv/AttributeKeyTemplate.java
jack-berg: I agree. We don't normalize attribute keys elsewhere, and this shouldn't be the exception.

src/main/java/io/opentelemetry/semconv/AttributeKeyTemplate.java
AlexanderWert: I'm fine with any of the above proposed approaches.

> I suppose using a single map minimizes the memory footprint of initializing many maps, most of which won't get used for a particular application. But we could solve that by initializing the maps with a small initial capacity.

Another alternative would be to do lazy initialization of the individual maps, to reduce the impact of template attributes that are not being used.

src/main/java/io/opentelemetry/semconv/AttributeKeyTemplate.java
```diff
@@ -14,9 +14,9 @@
 
 public final class AttributeKeyTemplate<T> {
 
-  private final ConcurrentMap<String, AttributeKey<T>> keysCache = new ConcurrentHashMap<>();
   private final String prefix;
   private final Function<String, AttributeKey<T>> keyBuilder;
+  private ConcurrentMap<String, AttributeKey<T>> keysCache;
```
trask, line 19: I believe this would need to be `volatile` for thread safety. I'd suggest staying with `final` and inline initialization.

src/main/java/io/opentelemetry/semconv/AttributeKeyTemplate.java
trask: > With a single map/cache we would need to cast the AttributeKey objects, though, when retrieving.

this seems like a good enough reason to stick with instance field

src/main/java/io/opentelemetry/semconv/AttributeKeyTemplate.java
jack-berg: Technically I think this is one of those weird situations where it doesn't actually matter if two threads read this as null and instantiate / use different instances of `ConcurrenHashMap`, since the consequence is just that you will have created two copies of the same `AttributeKey` instead of returning a cached copy.

The inline initialization is simpler though and doesn't raise flags about thread safety. I'd say stick with the `final` inline initialization and keep this type of thing in our back pocket if we get feedback about utilization.

AlexanderWert:
@jack-berg , @trask , @mateuszrzeszutek , @breedx-splk 

Based on the discussions above I reverted the implementation to use a final , inline-initialized map (with initial size 1) again. 
PTA(nother)L, Thanks!

src/main/java/io/opentelemetry/semconv/AttributeKeyTemplate.java
mateuszrzeszutek: > > With a single map/cache we would need to cast the AttributeKey objects, though, when retrieving.
> 
> this seems like a decent reason to stick with instance field

That's a good point. I don't have strong feelings about that, I think the current version is okay too.

src/main/java/io/opentelemetry/semconv/AttributeKeyTemplate.java
```diff
@@ -0,0 +1,78 @@
...
+package io.opentelemetry.semconv;
+
+import io.opentelemetry.api.common.AttributeKey;
+import java.util.List;
+import java.util.Locale;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.function.Function;
+
+public final class AttributeKeyTemplate<T> {
```
mateuszrzeszutek, line 15: Can you add JavaDoc on the class level too?

src/main/java/io/opentelemetry/semconv/AttributeKeyTemplate.java
```diff
@@ -0,0 +1,78 @@
...
+  static AttributeKeyTemplate<Double> doubleKeyTemplate(String prefix) {
+    return new AttributeKeyTemplate<Double>(prefix, AttributeKey::doubleKey);
+  }
+
+  static AttributeKeyTemplate<List<Double>> doubleArrayKeyTemplate(String prefix) {
+    return new AttributeKeyTemplate<List<Double>>(prefix, AttributeKey::doubleArrayKey);
+  }
+
+  private AttributeKey<T> createAttributeKey(String keyName) {
+    String key = prefix + "." + keyName.toLowerCase(Locale.ROOT);
```
mateuszrzeszutek, line 59: I have somewhat mixed feelings about lowercasing the key here. We did that for the HTTP headers, but there were some discussions about whether we should keep or ignore the case; and maybe, for other template attributes in the future we'll want to keep the case. I find that unlikely though -- I'm guessing we'll probably follow the precedent set by HTTP headers; and `key` and `Key` being different things is _super_ confusing.

mateuszrzeszutek (approved):
LGTM, I left one more minor JavaDoc comment.
Thanks!

src/main/java/io/opentelemetry/semconv/AttributeKeyTemplate.java
AlexanderWert: Yeah, I was basically following the HTTP headers example. But agree that for others we might need to keep the key as is. What do others think?

AlexanderWert: Actually let me change that to not lowercasing it here and do the `toLowerCase()` in the instrumentation when using it for HTTP headers. WDYT?

mateuszrzeszutek: > Actually let me change that to not lowercasing it here and do the `toLowerCase()` in the instrumentation when using it for HTTP headers. WDYT?

👍 
Should be a bit more future-proof in case the semconv decides to use case sensitive naming somewhere (I hope they don't 😅 )

AlexanderWert:
OK, then I think it's ready to be merged.

src/main/java/io/opentelemetry/semconv/AttributeKeyTemplate.java
```diff
@@ -0,0 +1,87 @@
...
+  private final String prefix;
+  private final Function<String, AttributeKey<T>> keyBuilder;
+  private final ConcurrentMap<String, AttributeKey<T>> keysCache = new ConcurrentHashMap<>(1);
+
+  AttributeKeyTemplate(String prefix, Function<String, AttributeKey<T>> keyBuilder) {
+    this.prefix = prefix;
+    this.keyBuilder = keyBuilder;
+  }
+
+  static AttributeKeyTemplate<String> stringKeyTemplate(String prefix) {
```
jack-berg, line 35:
```suggestion
  public static AttributeKeyTemplate<String> stringKeyTemplate(String prefix) {
```
What do folks think about making these public? As is, they can only be used for attributes in `ResourceAttributes` and `SemanticAttributes`, but its reasonable for a user to want to define their own custom attribute template. Maybe that suggests that ultimately this class should be promoted to `io.opentelemetry.api.common` to live alongside `AttributeKey` and `Attributes`.

jack-berg (approved):
Couple nits and a question about whether the methods to create `AttributeKeyTemplate` should be public to allow users to create their own templates. None of the comments should block this PR.

PR #35, opened 2023-11-14
Semconv 1.23.0

jack-berg:
There is some interaction between the changes in `semantic-conventions` v1.23.0 (with several groups of attributes moving to the registry) and build-tools `--only resource` argument. In a nutshell, several groups of resource attributes are not generated. We'll need to find a fix before proceeding with the 1.23.0 release of this repo.

Leaving this PR as a draft to demonstrate the issue.


jack-berg:
Closing in favor of #36.

Issue #6, opened 2023-08-11
Develop strategy for deprecating / removing experimental semantic conventions


jack-berg:
We currently maintain many outdated attributes with `@Deprecated` annotations and previously agreed that this is unsustainable.

For stable semantic conventions, we should obviously retain attributes indefinitely, even if they undergo allowed renames via schema transformation rules.

However, for experimental conventions, we should develop some procedure for end-of-lifing attributes as the conventions churn.

Once we develop this strategy, we should consider removing all current attributes marked as deprecated.

breedx-splk:
I think as long as the prior `@Deprecated` version indicates what the new thing is called or what the user should use instead, then it should be safe to remove these.

Can't the "procedure" to remove these just be a simple as removing them after a release cycle has passed?

lmolkova:
Created generalized issue in the semconv repo - https://github.com/open-telemetry/semantic-conventions/issues/550
Python has the same problem and it's be great to have language-agnostic solution

jack-berg:
This is resolved. See https://github.com/open-telemetry/semantic-conventions-java/issues/50#issuecomment-1998332270 for a full summary of the strategy.

PR #36, opened 2023-11-17
Semconv 1.23.1

jack-berg:
Replacement for #35.


jack-berg:
We should delay merging / releasing this until #24 is merged.

buildscripts/templates/SemanticAttributes.java.j2
```diff
@@ -97,6 +97,30 @@ public final class {{class}} {
 
     {%- endfor %}
 
+    {%- if class_name == "MessagingOperationValues" %}
```
jack-berg, line 100: I don't see any evidence in the template that we've accommodated the removal of values from semantic conventions enums, which manifest as static inner classes. But this seemed like the most correct thing to do. Also, we're using japicmp to ensure we don't make incompatible changes (until we resolve #6) and not fixing this causes japicmp to fail.

mateuszrzeszutek (approved):

jack-berg:
I've merged #24 and I think we're good to merge this and cut a release of 1.23.1.

PR #40, opened 2023-12-02
Semconv code generation prototype

lmolkova:
This is a prototype (related to https://github.com/open-telemetry/semantic-conventions/issues/551) of new code-gen approach:
- generates stable attributes in `io.opentelemetry.semconv`
- generates experimental attributes in `io.opentelemetry.semconv.vA_B_C`
- groups attributes by root namespace and generates them in multiple files

The process of codegen might looks like:
- keep old `Semantic|ResourceAttributes` with all deprecated things, delete at some point in the future
- update the version
- generate a new folder for experimental, update stable

As a result, we only need to care about back-compat when a stable attribute changes in a breaking manner.
Multiple instrumentation libs can send different semconv versions at once.

**Bloating: is indeed a problem**
- the size of the artifact produced with this change is 131 KB
- if I generate a prev version -  v1.22.0, it becomes 202 KB
 
TODO:
- [x] Figure out how to populate schema URL 
  - manually for now (with automated check)
  - not possible(?) with jinja, need some custom script to add new value to Java file
  - [ ] do the script
- [x] stable and deprecated are  mutually exclusive now, so template would not work in the future if a stable attribute gets deprecated - https://github.com/open-telemetry/semantic-conventions/issues/582


src/main/java/io/opentelemetry/semconv/UserAgentAttributes.java
```diff
@@ -0,0 +1,25 @@
...
+// DO NOT EDIT, this is an Auto-generated file from
+// buildscripts/templates/SemanticAttributes.java.j2
+@SuppressWarnings("unused")
+public final class UserAgentAttributes {
+
+  /**
+   * Value of the <a href="https://www.rfc-editor.org/rfc/rfc9110.html#field.user-agent">HTTP
+   * User-Agent</a> header sent by the client.
+   */
+  public static final AttributeKey<String> USER_AGENT_ORIGINAL = stringKey("user_agent.original");
```
lmolkova, line 21: Q: do we want to have `UserAgentAttributes.USER_AGENT_ORIGINAL` or `UserAgentAttributes.ORIGINAL` ?

trask (commented):
I have a slight preference for having two separate artifacts, one for stable attributes and one for (the last 5) experimental namespaced attributes

build.gradle.kts
```diff
@@ -124,19 +125,19 @@ val generateResourceAttributes by tasks.registering(Exec::class) {
...
-    "-v", "$projectDir/src/main/java/io/opentelemetry/semconv/:/output",
-    "otel/semconvgen:$generatorVersion",
-    "--only", "resource",
+    "-v", "$projectDir/src/main/java/io/opentelemetry/semconv/$semanticConventionsVersionEscaped/:/output",
+    "semconvgen:$generatorVersion",
     "--yaml-root", "/source", "code",
     "--template", "/templates/SemanticAttributes.java.j2",
-    "--output", "/output/ResourceAttributes.java",
-    "-Dclass=ResourceAttributes",
+    "--output", "/output/Attributes.java",
```
lmolkova, line 132: `/optput/Attributes.java` is actually a template and will be transformed into `/output/{RootNamespace}Attributes.java` - do we want to have more flexibility and have an actual template here like (`/output/SometingElse{RootNamespace}Attributes.java`)

src/main/java/io/opentelemetry/semconv/HttpAttributes.java
```diff
@@ -0,0 +1,159 @@
...
+   *       captured. Including all request headers can be a security risk - explicit configuration
+   *       helps avoid leaking sensitive information. The {@code User-Agent} header is already
+   *       captured in the {@code user_agent.original} attribute. Users MAY explicitly configure
+   *       instrumentations to capture them even though it is not recommended. The attribute value
+   *       MUST consist of either multiple header values as an array of strings or a single-item
+   *       array containing a possibly comma-concatenated string, depending on the way the HTTP
+   *       library provides access to headers.
+   * </ul>
+   */
+  public static final AttributeKeyTemplate<List<String>> HTTP_REQUEST_HEADER =
```
lmolkova, line 37: Q: do we want attribute and metric definitions to be in the same file or different files?

1. `HttpAttributes.HTTP_REQUEST_HEADER` (or `HttpAttributes.REQUEST_HEADER`) and `HttpMetrics.CLIENT_REQUEST_DURATION`

2. `Http.REQUEST_HEADER_ATTRIBUTE` and `Http.CLIENT_REQUEST_DURATION_METRIC`

trask: > 1. `HttpAttributes.HTTP_REQUEST_HEADER`  and `HttpMetrics.HTTP_CLIENT_REQUEST_DURATION`

I like this option 👍

src/main/java/io/opentelemetry/semconv/UserAgentAttributes.java
trask: > UserAgentAttributes.USER_AGENT_ORIGINAL

this is nice for statically importing the constants

though @mateuszrzeszutek recently said in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/9925#discussion_r1407425481

> static imports made this PR harder to write. If we ever decide to change things in the future again, I think I'd prefer to not have static imports here.

build.gradle.kts
lmolkova: resolving it for now as not substantial - it can always be done later

src/main/java/io/opentelemetry/semconv/v1_23_1/CloudeventsAttributes.java
```diff
@@ -0,0 +1,58 @@
...
+package io.opentelemetry.semconv.v1_23_1;
+
+import static io.opentelemetry.api.common.AttributeKey.stringKey;
+
+import io.opentelemetry.api.common.AttributeKey;
+
+// DO NOT EDIT, this is an Auto-generated file from
+// buildscripts/templates/SemanticAttributes.java.j2
+@SuppressWarnings("unused")
+public final class CloudeventsAttributes {
```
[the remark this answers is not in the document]
lmolkova: TODO: add custom overrides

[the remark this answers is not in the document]
jack-berg: I'm not sure about custom overrides. We currently translate `snake_case` to `UpperCamel`, and the problem here is that we are presented with a value that is normally thought of as two words subject to camel case (i.e. `CloudEvents`) as a single word (i.e. `cloudevents`).

If we add a set of special cases, to replace `cloudevents` with `CloudEvents`, then we susceptible to conflicts in the future if a word comes along which is matches our override. I.e. if later `cloud_events` comes along, then both `cloud_events` and `cloudevents` map to `CloudEvents`. 

Better I think to not have special cases, and fix upstream in semantic conventions in obvious cases.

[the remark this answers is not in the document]
lmolkova: created https://github.com/open-telemetry/semantic-conventions/issues/599

jack-berg:
Closing as this is superseded by #45.

PR #41, opened 2023-12-12
Split out incubating artifact

### LINKED ISSUE #22: Develop strategy for differentiating stable vs. experimental conventions, publishing stable artifact
The semantic conventions contain a mix of stable and experimental conventions. This maturity level is denoted in the generated markdown (e.g. [see stable resource conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/README.md#service)), but not in the yaml representation used to generated java classes (e.g. see [resource conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/model/resource/service.yaml)).

Semantic conventions will always be a work in progress to some extent, with some domains / attributes marked as experimental. We need a strategy for being able to release the stable conventions in a stable way with backwards compatibility guarantees, while also allowing for experimental conventions to evolve and break over time.

jack-berg:
Resolves #22. Embodies this [proposal](https://github.com/open-telemetry/semantic-conventions-java/issues/22#issuecomment-1850770122). Uses some of the concepts in #40, but without breaking out classes for each top level namespace.

Summary:
- Break out new `io.opentelemetry.semconv:opentelemetry-semconv-incubating` artifact for experimental semantic conventions.
- Incubating artifact has files `IncubatingResourceAttributes` and `IncubatingSemanticAttributes`. (I think breaking out classes for top level namespaces in #40 is a good thing to do but can be done separately.)
- Delete all hardcoded deprecated constants from jinja template. We've been keeping those around while we work out a strategy for #22, and we now have a strategy. Additionally, its not obvious where we would put them. They don't belong in the artifact for stable conventions, and the names of the classes changes in the incubating artifact, so would break users anyway. Furthermore, I think we should break out classes for top level namespaces in a followup PR, which makes it harder to justify including the old deprecated constants.
- Add new SchemaUrls class with constants of form `public static final String V1_23_1 = "https://opentelemetry.io/schemas/1.23.1"`. This is a running list of schema URL constants which needs to be manually updated.


jack-berg:
I pushed a commit that allows the same thing to be accomplished without https://github.com/open-telemetry/build-tools/pull/243.

Its less convenient, but allows this PR to proceed independently of #40. Marking as ready to review.

README.md
```diff
@@ -15,11 +15,18 @@ Published releases are available on maven central. Replace `{{version}}` with th
...
+    <!-- Stable semantic conventions. Note: generated code is still subject to breaking changes while published with "-alpha" suffix. -->
     <dependency>
       <groupId>io.opentelemetry.semconv</groupId>
       <artifactId>opentelemetry-semconv</artifactId>
       <version>{{version}}</version>
     </dependency>
+    <!-- Incubating semantic conventions. Breaking changes expected. Library instrumentation SHOULD NOT depend on this. -->
+    <dependency>
+      <groupId>io.opentelemetry.semconv</groupId>
+      <artifactId>opentelemetry-semconv.incubating</artifactId>
```
trask, line 27:
```suggestion
      <artifactId>opentelemetry-semconv-incubating</artifactId>
```

build.gradle.kts
```diff
@@ -54,25 +47,10 @@ nexusPublishing {
...
-  testImplementation("org.junit.jupiter:junit-jupiter-engine")
-
-  testImplementation(platform("org.assertj:assertj-bom:3.24.2"))
-  testImplementation("org.assertj:assertj-core")
-}
-
 // start - define tasks to download, unzip, and generate from opentelemetry/semantic-conventions
 var generatorVersion = "0.23.0"
 val semanticConventionsRepoZip = "https://github.com/open-telemetry/semantic-conventions/archive/v$semanticConventionsVersion.zip"
+// val semanticConventionsRepoZip = "https://github.com/lmolkova/semantic-conventions/archive/bd833aa5ff13996293654dc45a4a61354b0bbe5d.zip"
```
trask, line 53:
```suggestion
```

trask (approved):
👍

PR #45, opened 2024-02-09
Generate classes by root namespace

jack-berg:
A rebase of #40 on top of #41. @lmolkova it was impractical to do a traditional merge because there were just too many conflicts, so I listed you as a co-author on the commit. Hope that's ok with you.

This PR breaks out classes for each top level namespace in semantic conventions:
- For stable conventions, classes are named `{{Namespace}}Attributes.java` 
- For incubating / experimental conventions, classes are named `{{Namespace}}IncubatingAttributes.java`

This omits the automatic metric generation of #40. Figure we can do that in a separate PR to keep changes organized.

Leaving this as a draft because we're waiting https://github.com/open-telemetry/build-tools/pull/243, which is currently only merged to the `feature/codegen-by-namespace` branch, which doesn't have an image published yet. I think https://github.com/open-telemetry/build-tools/pull/256 might be trying to solve that.


build.gradle.kts
```diff
@@ -48,7 +48,9 @@ nexusPublishing {
 }
 
 // start - define tasks to download, unzip, and generate from opentelemetry/semantic-conventions
-var generatorVersion = "0.23.0"
+// TODO: build locally from https://github.com/open-telemetry/build-tools/tree/feature/codegen-by-namespace branch
+// Blocked on release of this: https://github.com/open-telemetry/build-tools/pull/243#issuecomment-1936446767
+var generatorVersion = "codegen-by-namespace"
```
arminru, line 53:
```suggestion
var generatorVersion = "feature-codegen-by-namespace"
```
see https://hub.docker.com/r/otel/semconvgen/tags

jack-berg: 🎉

jack-berg:
Update: A image has been published containing the logic form the `feature/codegen-by-namespace` branch. Marking this as ready for review since the new capabilities are expected to eventually merge into `main`.

trask (commented):
Do we know that this is the direction other languages are going to take? I like it, my only reservation is to minimize the number of times we change the structure before marking it stable.

jack-berg:
We can't say for sure, but this [issue](https://github.com/open-telemetry/semantic-conventions/issues/551) contains feedback from a number of language maintainers which is telling:
- [js maintainer](https://github.com/open-telemetry/semantic-conventions/issues/551#issuecomment-1827934380) seems to indicate that grouping attributes by namespace isn't very important
- [dotnet maintainers](https://github.com/open-telemetry/semantic-conventions/issues/551#issuecomment-1830365716) say grouping attributes by namespace makes sense
- [go maintainer](https://github.com/open-telemetry/semantic-conventions/issues/551#issuecomment-18344299620) says "is desirable to have the semantic conventions laid out in files that matched their groups so we can find them easier". I think this is referring to grouping by namespace?
- [c++ maintainer](https://github.com/open-telemetry/semantic-conventions/issues/551#issuecomment-1854829955) says group by namespace sounds good
- [erlang maintainer](https://github.com/open-telemetry/semantic-conventions/issues/551#issuecomment-1863607296) agrees with c++ maintainer
- [swift](https://github.com/open-telemetry/semantic-conventions/issues/551#issuecomment-1845983679) and [php](https://github.com/open-telemetry/semantic-conventions/issues/551#issuecomment-1863620858) don't seem to have a preference on grouping

Didn't here from all maintainers, but there seems to agreement about grouping by namespace amongst maintainers who expressed an opinion.

Not related to this comment, but related to this PR: This [issue](https://github.com/open-telemetry/semantic-conventions/issues/740) to have separate properties from stability and deprecation, and not remove deprecated attributes will be something we will want to take advantage of.

lmolkova:
> Do we know that this is the direction other languages are going to take? I like it, my only reservation is to minimize the number of times we change the structure before marking it stable.

Thanks @jack-berg for summarizing the feedback!
I was also checking with python SIG - they are fine with per-root-namespace approach https://github.com/open-telemetry/opentelemetry-python/pull/3586

@trask I only know about two possible directions: 
- current (file per signal or similar)
- group by root namespace - this is a new feature in the build-tools

I think there could be something else, but nobody requests for it explicitly and nobody works on supporting it (build-tools changes would be necessary).

So at least for today, there is no third way.

jsuereth:
FYI - This PR is broken against latest feature head.  You need to merge this: https://github.com/jack-berg/semantic-conventions-java/pull/1

jack-berg:
@open-telemetry/java-approvers - WDYT? Are there any other questions / concerns about moving forward with this?

trask:
I didn't see any deprecated attributes (yet?), is this expected?

trask (approved):
heads up @brunobat

jack-berg:
> I didn't see any deprecated attributes (yet?), is this expected?

@trask good observation. I wouldn't expect to see any for 1.23.1, but I also don't see any in #46, which upgrades to 1.24.0 which splits out the deprecated and stability properties. I think I found the [issue](https://github.com/open-telemetry/build-tools/pull/244#issuecomment-1971418118).

lmolkova:
@jack-berg @trask I'm in doubt how  to approach stabilization and wonder if you have some thoughts on it

E.g.
1. attr `foo.bar` goes stable
2. it disappears from incubating 
3. it appears in stable

Assuming application uses `semconv.incubating.FooAttributes.FOO_BAR` it will be a breaking change (or if instrumentation lib version is not aligned).

I wonder if we can do a better job if incubating artifact included all attributes - stable and experimental - then experimental-> stable transition would not be breaking.
When attribute becomes stable we could also mark corresponding constant as `@Deprecated - use  `semconv.FooAttributes.FOO_BAR` instead`

buildscripts/templates/SemanticAttributes.java.j2
```diff
@@ -31,18 +25,18 @@
   {%- elif type == "double" -%}
     doubleKey
   {%- else -%}
-    {{lowerFirst(type)}}Key
+    {{type | to_camelcase(False)}}Key
   {%- endif -%}
 {%- endmacro %}
 {%- macro print_value(type, value) -%}
```
lmolkova, line 31: nit: you can move it and use built-in 

``{{ attribute | print_member_value(member) }}``

instead.

see https://github.com/open-telemetry/opentelemetry-python/blob/cbebd89c6da683c0516665a929774062ae93b42d/scripts/semconv/templates/semantic_attributes.j2#L50 for full example

buildscripts/templates/SemanticAttributes.java.j2
```diff
@@ -31,18 +25,18 @@
...
   {{ "\"" if type == "String"}}{{value}}{{ "\"" if type == "String"}}
 {%- endmacro %}
-{%- macro upFirst(text) -%}
-  {{ text[0]|upper}}{{text[1:] }}
-{%- endmacro %}
-{%- macro lowerFirst(text) -%}
-  {{ text[0]|lower}}{{text[1:] }}
-{%- endmacro %}
+{%- set filtered_attributes = attributes_and_templates | select(filter) | list %}
```
lmolkova, line 34: you might want to exclude certain namespaces (e.g. `aspnetcore` is not particularly helpful on Java)

Here's a basic [example](https://github.com/open-telemetry/opentelemetry-python/blob/cbebd89c6da683c0516665a929774062ae93b42d/scripts/semconv/templates/semantic_attributes.j2#L1)

jack-berg:
Ok, I've pushed a commit to use the latest release of the `feature/codegen-by-namespace`, which adds a [--strict-validation](https://github.com/open-telemetry/build-tools/pull/281). This allows it to work with 1.23.1 of semconv which didn't separate out the stability and deprecated properties. 

The result is that we now see deprecated attributes, which addresses @trask [comment](https://github.com/open-telemetry/semantic-conventions-java/pull/45#issuecomment-1971423570). However, the deprecated annotations are missing the `@Deprecated` annotation, because 1.23.1 of semconv conflates stability and deprecated in a single property. This is resolved once we bump to 1.24.0 of semconv (#46). See [the diff](https://github.com/jack-berg/semantic-conventions-java/compare/namespaced-attributes...jack-berg:semantic-conventions-java:semconv-1.24.0) for properly working `@Deprecated` annotations.

jack-berg:
> I wonder if we can do a better job if incubating artifact included all attributes - stable and experimental - then experimental-> stable transition would not be breaking.
When attribute becomes stable we could also mark corresponding constant as @Deprecated - use  semconv.FooAttributes.FOO_BAR instead

I think that could work. Incubating is then a superset of stable.

buildscripts/templates/SemanticAttributes.java.j2
```diff
@@ -74,42 +68,23 @@ public final class {{class}} {
         <ul> {{attribute.note | replace("> ", "") | render_markdown(code="{{@code {0}}}", paragraph="<li>{0}</li>", list="{0}")}} </ul>
 
     {%- endif %}
-    {%- if (attribute.stability | string())  == "StabilityLevel.DEPRECATED" %}
+    {%- if attribute | is_deprecated %}
   *
   * @deprecated {{attribute.brief | to_doc_brief}}.
```
lmolkova, line 73: I *think* this should fix deprecated annotations and work for both 1.23.1 and 1.24.0

```suggestion
  * @deprecated {{ attribute.deprecated if attribute.deprecated else attribute.brief | to_doc_brief}}.
```

lmolkova: ah, wait, the problem is in `attribute | is_deprecated`, so  to make it work with 1.23.x and 1.24.0 would be to change it to `if attribute.deprecated` (`deprecated` is a string)

lmolkova:
> > I wonder if we can do a better job if incubating artifact included all attributes - stable and experimental - then experimental-> stable transition would not be breaking.
> > When attribute becomes stable we could also mark corresponding constant as @deprecated - use  semconv.FooAttributes.FOO_BAR instead
> 
> I think that could work. Incubating is then a superset of stable.

Tried it here - https://github.com/jack-berg/semantic-conventions-java/pull/2
Generation of @deprecated annotation is a bit ugly, but possible.

buildscripts/templates/SemanticAttributes.java.j2
jack-berg: Agreed. I opened a PR but think we should address this separately.

jack-berg:
@lmolkova I've merged your PR and an additional commit that allows this to work as is. I think we should be good to merge this PR now. 

@trask let me know if you have any additional thoughts.

lmolkova:
> @lmolkova I've merged your PR and an additional commit that allows this to work as is. I think we should be good to merge this PR now.
> 
> @trask let me know if you have any additional thoughts.

can you please revert it? It relies on unmerged changes in the build-tools that are unlikely to make it into the next build-tools release.  I don't think it's blocking since I assume it should be fine to add a bunch or stable attributes into incubating artifact later on. 

In the meantime, I really want  to get your thoughts on  how this stable+unstable solution might look like.

Repeating the question I left on the python PR https://github.com/open-telemetry/opentelemetry-python/pull/3586#issuecomment-1973978521

**Option 1**: ship 2 artifacts  (`semconv` and `semconv-incubating`) - current solution

It's aligned with semantic versioning, easy to maintain. 
Still having stable attributes as deprecated in the incubating artifact is not great.

**Option 2**: ship 1 artifact, but two different versions: stable and preview. Stable contains only stable semconv. Preview contains all of them.

Ideal in terms of semantic versioning. When attribute stabilizes, it remains in the same namespace and all libs/apps that depend on it don't need to change
The only downside is that it maybe harder to maintain and release.

**Option 3**: Probably not applicable to Java - ship 1 artifact, 1 version containing stable and experimental semconv. Emphasize that some parts are experimental via namespace (`import semconv.experimental.foo_attributes`, perhaps we should call it `_experimental` )?

It's easy to do, but violates semantic versioning. When attribute stabilizes, it moves to a different namespace, i.e. libs/apps that depend on it need to change.

----------------------------
I think the cleanest solution from user perspective would be option 2:

we'd have `semantic-conventions-java:1.0.0` with stable only, `semantic-conventions-java:1.0.0-beta.1` with all semconvs. Experimental -> stable transition requires no action on the user side. 
It's also one of the common patters to ship experimental features.

We can also discuss at the SIG meeting,  but I always have a conflict discussing the same with Python SIG at the same time:)

jack-berg:
@lmolkova I adjusted it so it only uses build-tools code merged to the feature branch which is published to docker hub

trask:
I'd suggest that we wait on merging this (or at least block on releasing it) until we have final decision on #50, so that we can limit the breaking changes to a single release

jack-berg:
I think we should hold the release until we resolve #50, but we've already merged substantial breaking changes to `main`, so I don't see the value of holding off on this specific PR.

trask (approved):

PR #46, opened 2024-02-09
Semconv 1.24.0

jack-berg:
Depends on #45.


jack-berg:
#45 is merged so this is ready for review.

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/EventIncubatingAttributes.java
```diff
@@ -15,34 +15,19 @@
...
-   *   <li>Events across different domains may have same {@code event.name}, yet be unrelated
-   *       events.
+   *   <li>Event names are subject to the same rules as <a
+   *       href="https://github.com/open-telemetry/opentelemetry-specification/tree/v1.26.0/specification/common/attribute-naming.md">attribute
+   *       names</a>. Notably, event names are namespaced to avoid collisions and provide a clean
+   *       separation of semantics for events in separate domains like browser, mobile, and
+   *       kubernetes.
    * </ul>
    */
-  public static final AttributeKey<String> EVENT_DOMAIN = stringKey("event.domain");
```
jack-berg, line 27: @lmolkova This is an example of an experimental attribute that was deleted rather retained but marked as deprecated.

Based on what you said, in the future this will not be the case. And we would expect to keep this attribute around semi-permanently marked as deprecated. Is that correct?

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/AspnetcoreIncubatingAttributes.java
```diff
@@ -0,0 +1,97 @@
...
+
+import static io.opentelemetry.api.common.AttributeKey.booleanKey;
+import static io.opentelemetry.api.common.AttributeKey.stringKey;
+
+import io.opentelemetry.api.common.AttributeKey;
+
+// DO NOT EDIT, this is an Auto-generated file from
+// buildscripts/templates/SemanticAttributes.java.j2
+@SuppressWarnings("unused")
+public final class AspnetcoreIncubatingAttributes {
```
jack-berg, line 16: Add to exclude list once #54 is merged.

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/SignalrIncubatingAttributes.java
```diff
@@ -0,0 +1,58 @@
...
+package io.opentelemetry.semconv.incubating;
+
+import static io.opentelemetry.api.common.AttributeKey.stringKey;
+
+import io.opentelemetry.api.common.AttributeKey;
+
+// DO NOT EDIT, this is an Auto-generated file from
+// buildscripts/templates/SemanticAttributes.java.j2
+@SuppressWarnings("unused")
+public final class SignalrIncubatingAttributes {
```
jack-berg, line 15: Add to exclude list once #54 is merged.

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/AspnetcoreIncubatingAttributes.java
breedx-splk: It's approved. Don't forget. 🙃

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/EventIncubatingAttributes.java
breedx-splk: I'm not trying to answer for @lmolkova, and I'm not addressing the general question being asked....but in this specific case I do think it makes sense to remove it, because (experimental) Events are starting to gel and we don't want people using this.

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/HttpIncubatingAttributes.java
```diff
@@ -112,7 +127,12 @@ public final class HttpIncubatingAttributes {
   public static final AttributeKey<Long> HTTP_REQUEST_RESEND_COUNT =
       longKey("http.request.resend_count");
 
-  /** Deprecated, use {@code http.request.header.content-length} instead. */
+  /**
+   * Deprecated, use {@code http.request.header.content-length} instead.
+   *
+   * @deprecated Deprecated, use `http.request.header.content-length` instead.
```
breedx-splk, line 133: This isn't a showstopper, but it does really a little redundant.

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/SystemIncubatingAttributes.java
```diff
@@ -25,10 +25,6 @@ public final class SystemIncubatingAttributes {
   /** The device identifier */
   public static final AttributeKey<String> SYSTEM_DEVICE = stringKey("system.device");
 
-  /** The disk operation direction */
-  public static final AttributeKey<String> SYSTEM_DISK_DIRECTION =
```
breedx-splk, line 29: This might be another set of examples of deleted semconv?

breedx-splk (approved):
These are always weird to review, because it's mostly generated code right? In any case, I did review it, seems ok to me.

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/EventIncubatingAttributes.java
lmolkova: sorry, I missed it.
Yes, starting from 1.24.0 we'll have a check that prevents attributes (and metrics) from being deleted - https://github.com/open-telemetry/semantic-conventions/pull/764 

It would apply to experimental and stable semconv. 

I think we'll need to do the same for events, updated https://github.com/open-telemetry/build-tools/issues/292  so we don't forget

breedx-splk: Thanks, I guess that's fine, as long as they're marked deprecated and can be removed in the subsequent release. I'd hate for us to have constraints that cause long-term maintenance burden due to changes in experimental semconv (and yes, I realize this isn't the place to drag this out, heh 😅 ).

semconv/src/main/java/io/opentelemetry/semconv/JvmAttributes.java
trask, line 1: 🎉

trask (approved):

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/EventIncubatingAttributes.java
lmolkova: Would love to hear your thoughts on removing-experimental-deprecated on https://github.com/open-telemetry/semantic-conventions/issues/740. It'd be great to have a common policy for all language SIGs.

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/HttpIncubatingAttributes.java
jack-berg: Good point, but we don't control that text. It comes directly from the semantic conventions YAML: https://github.com/open-telemetry/semantic-conventions/blob/cafda7127683b7f667e27cdbd3220510b6f998c9/model/registry/deprecated/http.yaml#L34

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/SystemIncubatingAttributes.java
jack-berg: Yeah I think the process to switch from deleting to deprecating and retaining forever (or at least a while) is a relatively new thing. We just started to see it in 1.24.0, and I expect it will be the norm going forward.

@lmolkova / @trask may be able to confirm.

jack-berg:
> These are always weird to review, because it's mostly generated code right? In any case, I did review it, seems ok to me.

Its all generated code. To some extent you can rely on the build passing to help give you confidence in the output, since the build will run code generation and detect any difference. The one thing it doesn't check is if there are are files which are part of the commit, but which are not generated anymore. For example, this [commit](https://github.com/open-telemetry/semantic-conventions-java/pull/46/commits/7dcb7e31bb409cb5916f10d7d0ba84a89a8a12de) excludes the `aspnetcore` and `signlar` root namespaces, but the build would have passed even had I not manually deleted the `AspnetcoreIncubatingAttributes` and `SignalrIncubatingAttributes` files.

PR #52, opened 2024-03-14
Cleanup enum generation

### LINKED ISSUE #49: Generate javadoc / deprecated annotations for enum definitions
I noticed that our enum definitions don't have javadoc or deprecated annotations (when applicable). See [example](https://github.com/open-telemetry/semantic-conventions-java/blob/0141bda53279d6a1ac975bb05d20688f54368f01/semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/IncubatingSemanticAttributes.java#L1947-L2000).

Let's fix this.

jack-berg:
Resolves #49.

- Adds enum javadoc pointing back to corresponding attribute constant
- Add deprecated javadoc / annotation based on same rules as attributes
- Exclude the `// Enum definitions` comment if a namespace doesn't have any enums

### DIFF
--- a/semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/HttpIncubatingAttributes.java
+++ b/semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/HttpIncubatingAttributes.java
@@ -193,6 +193,13 @@ public final class HttpIncubatingAttributes {
   public static final AttributeKey<String> HTTP_URL = stringKey("http.url");
 
   // Enum definitions
+  /**
+   * Values for {@link #HTTP_REQUEST_METHOD}.
+   *
+   * @deprecated deprecated in favor of stable {@link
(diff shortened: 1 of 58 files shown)


semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/HttpIncubatingAttributes.java
```diff
@@ -193,6 +193,13 @@ public final class HttpIncubatingAttributes {
   public static final AttributeKey<String> HTTP_URL = stringKey("http.url");
 
   // Enum definitions
+  /**
+   * Values for {@link #HTTP_REQUEST_METHOD}.
+   *
+   * @deprecated deprecated in favor of stable {@link
+   *     io.opentelemetry.semconv.HttpAttributes.HttpRequestMethodValues} attribute.
+   */
+  @Deprecated
```
jack-berg, line 202: This is a good example of the deprecated annotation. Javadoc links back to the stable equivalent.

trask (approved):

PR #54, opened 2024-03-15
Exclude namespaces that aren't useful in jvm environments

### LINKED ISSUE #48: Filter namespaces for non-java runtimes
you might want to exclude certain namespaces (e.g. `aspnetcore` is not particularly helpful on Java)

Here's a basic [example](https://github.com/open-telemetry/opentelemetry-python/blob/cbebd89c6da683c0516665a929774062ae93b42d/scripts/semconv/templates/semantic_attributes.j2#L1) I'm entertaining in python

_Originally posted by @lmolkova in https://github.com/open-telemetry/semantic-conventions-java/pull/45#discussion_r1509630367_

jack-berg:
Resolves #48.


breedx-splk (approved):
Yeah, this is nice.

trask (approved):
I'm not sure it's worth it, but no objection

PR #70, opened 2024-05-15
Use weaver for semantic convention codegen

jsuereth:
I wasn't sure what you might want to change/fix about codegen when moving to weaver, so this attempts to keep the status quo as best as possible.

## TODOs

- [x] Import markdown / notes -> javadoc filter to the Jinja weaver uses.
- [x] Finish supporting *all* types of template keys or throw an exception if an unsupported one shows up.
- [x] Remove previous jinja template
- [x] filter unwanted semconv groups in JQ expression in weaver config.
- [x] figure out what to do w/ K8S convention camel case.
- [x] Use type_mapping in weaver.yaml instead of macros.
- [x] Add `stability: deprecated` attributes back to incubating classes.

## Possible Improvements thanks to weaver

- We could generate *all* the code (for incubating + stable) in the same weaver config, with the same shared set of templates.  I kept it as-is for now so you can more easily compare what's happening.
- We can register name conversions for method/class/etc. for easier to understand case conversions.
- We can register acronymns, for consistent capitalization.


buildscripts/templates/registry/incubating_java/IncubatingSemanticAttributes.java.j2
```diff
@@ -0,0 +1,108 @@
...
+  {%- elif attribute.type == "template[string]" -%}
+    stringTemplate
+  {%- elif attribute.type == "template[string[]]" -%}
+    stringArrayKeyTemplate
+  {%- elif attribute.type is mapping -%}
+    stringKey
+  {%- else -%}
+    {{ type }}Key
+  {%- endif -%}
+{%- endmacro %}
```
lquerel, line 43: We could generalize the `type_mapping` section in weaver.yaml and support multiple mapping filters.
See https://github.com/open-telemetry/weaver/issues/156

jsuereth: I like this idea.

buildscripts/templates/registry/java/SemanticAttributes.java.j2
```diff
@@ -0,0 +1,108 @@
...
+    {# Note: Treat all enums as strings for now? #}
+    String
+  {%- elif attribute.type == "template[string]" -%}
+    String
+  {%- elif attribute.type == "template[string[]]" -%}
+    List<String>
+  {%- else -%}
+    {{ attribute.type }}
+  {%- endif -%}
+{%- endmacro %}
```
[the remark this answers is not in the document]
jsuereth: I'll add a TODO for that

buildscripts/templates/registry/incubating_java/IncubatingSemanticAttributes.java.j2
```diff
@@ -0,0 +1,108 @@
...
+    {# Note: Treat all enums as strings for now? #}
+    String
+  {%- elif attribute.type == "template[string]" -%}
+    String
+  {%- elif attribute.type == "template[string[]]" -%}
+    List<String>
+  {%- else -%}
+    {{ attribute.type }}
+  {%- endif -%}
+{%- endmacro %}
```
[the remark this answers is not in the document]
jsuereth: I was actually unable to, thanks to `enum` type.  PTAL.

buildscripts/templates/registry/incubating_java/IncubatingSemanticAttributes.java.j2
```diff
@@ -0,0 +1,108 @@
...
+
+import io.opentelemetry.api.common.AttributeKey;
+import io.opentelemetry.semconv.AttributeKeyTemplate;
+import java.util.List;
+
+// DO NOT EDIT, this is an Auto-generated file from buildscripts/templates/registry/java/SemanticAttributes.java.j2
+@SuppressWarnings("unused")
+public final class {{ my_class_name }} {
+    {% for attribute in ctx.attributes | sort(attribute="name") %}{% if attribute is experimental %}
+    {# TODO - We need the ability to convert markdown / text into javadoc safe strings. #}
```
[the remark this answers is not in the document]
jsuereth: I think we're still lacking some abiltiieshere.  Take a look at the diffs for notes.

buildscripts/templates/registry/java/SemanticAttributes.java.j2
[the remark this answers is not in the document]
jsuereth: Mostly fixed...

buildscripts/templates/registry/incubating_java/IncubatingSemanticAttributes.java.j2
jsuereth: I used this.  It still doesn't handle enums properly.

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/JvmIncubatingAttributes.java
```diff
@@ -1,157 +0,0 @@
...
-
-import static io.opentelemetry.api.common.AttributeKey.booleanKey;
-import static io.opentelemetry.api.common.AttributeKey.stringKey;
-
-import io.opentelemetry.api.common.AttributeKey;
-
-// DO NOT EDIT, this is an Auto-generated file from
-// buildscripts/templates/SemanticAttributes.java.j2
-@SuppressWarnings("unused")
-public final class JvmIncubatingAttributes {
```
jack-berg, line 16: Why are files like this missing missing?

semconv/src/main/java/io/opentelemetry/semconv/SchemaUrls.java
```diff
@@ -1,16 +0,0 @@
-/*
- * Copyright The OpenTelemetry Authors
- * SPDX-License-Identifier: Apache-2.0
- */
-
-package io.opentelemetry.semconv;
-
-public final class SchemaUrls {
```
jack-berg, line 8: This is not generated. We manually add constants to it with each release. There's a build check it make sure it has variables for each verion.

semconv/src/main/java/io/opentelemetry/semconv/SemanticAttributes.java
```diff
@@ -1,3910 +0,0 @@
...
-import java.util.List;
-
-/**
- * @deprecated This class is deprecated and will be removed in a future release. It is only provided
- *     as a convenience to help migration to the new semantic conventions classes structure that was
- *     introduced in version 1.24.0.
- */
-@Deprecated
-@SuppressWarnings("unused")
-public final class SemanticAttributes {
```
jack-berg, line 26: We're keeping a copy of this and `ResourceAttributes` around to ease the transition.

jack-berg (commented):
What's the difference between build tools and weaver? Is the community trying to transition away from build tools?

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/JvmIncubatingAttributes.java
jsuereth: This is now fixed

semconv/src/main/java/io/opentelemetry/semconv/SchemaUrls.java
jsuereth: Fixed

semconv/src/main/java/io/opentelemetry/semconv/SemanticAttributes.java
jsuereth: Fixed

jsuereth:
Update this, but we're now getting issues where weaver can't appropriately "javadoc-friendly" a comment, e.g.:

```
[ant:checkstyle] [ERROR] /home/joshuasuereth/src/open-telemetry/semantic-conventions-java/semconv/src/main/java/io/opentelemetry/semconv/ExceptionAttributes.java:26: <p> tag should be preceded with an empty line. [JavadocParagraph]
```

jsuereth:
@jack-berg This is ready for review now.  We have the javadoc/comment fixes in place for codegen.

I haven't consolidated down to calling weaver ONCE in gradle yet, but I can do that in a follow up PR.

buildscripts/templates/registry/incubating_java/IncubatingSemanticAttributes.java.j2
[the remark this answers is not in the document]
jsuereth: Fixed

buildscripts/templates/registry/java/weaver.yaml
```diff
@@ -0,0 +1,59 @@
+comment_formats:
```
jack-berg, line 1: Where can I read more about this file's contents?

jack-berg: https://github.com/open-telemetry/weaver/blob/main/docs/weaver-config.md

lquerel: And here https://github.com/open-telemetry/weaver/tree/main/crates/weaver_forge#comment-filter

build.gradle.kts
```diff
@@ -83,35 +83,36 @@ fun generateTask(taskName: String, incubating: Boolean) {
...
+      listOf()
+    else {
+      // Make sure we run as local file user
+      val unix = com.sun.security.auth.module.UnixSystem() 
+      val uid = unix.getUid() // $(id -u $USERNAME)
+      val gid = unix.getGid() // $(id -g $USERNAME)
+      listOf("-u", "$uid:$gid")
+    }
+    val weaver_args = listOf(
         "--rm",
```
jack-berg, line 100: Apparently the `otel/weaver` image doesn't publish for `arm64` platform yet. 
```suggestion
        "--rm",
        "--platform=linux/x86_64",
```

build.gradle.kts
```diff
@@ -83,35 +83,36 @@ fun generateTask(taskName: String, incubating: Boolean) {
...
+        "--registry=/home/weaver/source",
+        "--templates=/home/weaver/templates",
+        "$target",
+        "/home/weaver/target/")
+
+    setArgs(listOf("run") + file_args + weaver_args)
   }
 }
 
+// TODO - With weaver we can generate both of this in one go, but let's refactor smaller pieces at a time.
```
jack-berg, line 115: Meh.. we already have a `generateSemanticConventions` task which merged both into one. Also, it can be nice to have the ability to isolate to only generate stable or incubating.

buildscripts/checkstyle.xml
```diff
@@ -144,7 +144,9 @@
             <property name="forbiddenSummaryFragments"
              value="^@return the *|^This method returns |^A [{]@code [a-zA-Z0-9]+[}]( is a )"/>
         </module>
-        <module name="JavadocParagraph"/>
+        <module name="JavadocParagraph">
+            <property name="allowNewlineParagraph" value="false"/>
```
jack-berg, line 148: What's the motivation here?

jack-berg: Example of javadoc that fails the build without this. See the line with the lone `<p>` before the `<ul>` - weaver seems to use `<p> ` as a line break. 
```
public final class ErrorAttributes {
  /**
   * Describes a class of error the operation ended with.
   *
   * <p>Notes:
   *
   * <p>The {@code error.type} SHOULD be predictable, and SHOULD have low cardinality.
   *
   * <p>When {@code error.type} is set to a type (e.g., an exception type), its canonical class name
   * identifying the type within the artifact SHOULD be used.
   *
   * <p>Instrumentations SHOULD document the list of errors they report.
   *
   * <p>The cardinality of {@code error.type} within one instrumentation library SHOULD be low.
   * Telemetry consumers that aggregate data from multiple instrumentation libraries and
   * applications should be prepared for {@code error.type} to have high cardinality at query time
   * when no additional filters are applied.
   *
   * <p>If the operation has completed successfully, instrumentations SHOULD NOT set {@code
   * error.type}.
   *
   * <p>If a specific domain defines its own set of error identifiers (such as HTTP or gRPC status
   * codes), it's RECOMMENDED to:
   *
   * <p>
   *
   * <ul>
   *   <li>Use a domain-specific attribute
   *   <li>Set {@code error.type} to capture all errors, regardless of whether they are defined
   *       within the domain-specific set or not.
   * </ul>
   */
```

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/ContainerIncubatingAttributes.java
```diff
@@ -106,46 +91,11 @@ public final class ContainerIncubatingAttributes {
   public static final AttributeKeyTemplate<String> CONTAINER_LABEL =
       stringKeyTemplate("container.label");
 
-  /**
-   * Deprecated, use {@code container.label} instead.
-   *
-   * @deprecated Deprecated, use `container.label` instead.
-   */
-  @Deprecated
```
jack-berg, line 114: What's going on with the removal of these deprecated attributes?

build.gradle.kts
jsuereth: I can fix that, can you open a feature request on weaver?

build.gradle.kts
jsuereth: It just means you have to duplicate the weaver config yaml in both places for now

buildscripts/checkstyle.xml
jsuereth: Yes - that's the motivation. I asked on the one CNCF slack channel about this, but didn't hear any feedback.

I'm unable to figure out how to resolve this succesfully, weaver is using <p> and the actual underlying data in the YAML has this empty line break.

build.gradle.kts
jack-berg: Ah yeah would be good to DRY that up

buildscripts/checkstyle.xml
jack-berg: > I'm unable to figure out how to resolve this succesfully, weaver is using

Looks like some of the message got truncated

FWIW, I don't think its a big deal to have empty paragraphs.

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/ContainerIncubatingAttributes.java
jsuereth: I'll take a look.  Likely I've filtered them out in my JQ filter for some reason...

build.gradle.kts
jack-berg: https://github.com/open-telemetry/weaver/issues/345

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/ContainerIncubatingAttributes.java
jsuereth: It looks like all the experimental + deprecated attributes are not showing up here from the deprecated group in the registry, I'm investigating

buildscripts/templates/registry/incubating_java/weaver.yaml
```diff
@@ -0,0 +1,60 @@
...
+    trim: true
+    remove_trailing_dots: false
+default_comment_format: javadoc
+templates:
+  - pattern: IncubatingSemanticAttributes.java.j2
+    filter: > 
+      semconv_grouped_attributes({
+        "exclude_root_namespace": ["ios", "aspnetcore", "signalr"], 
+        "exclude_stability": ["deprecated"],
+        "exclude_deprecated": true
```
jsuereth, line 20: @jack-berg - This is why we don't see deprecated *and* experimental things.  Setting to false, they'll all come back.

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/ContainerIncubatingAttributes.java
jsuereth: Ah right - If you recall the `client.id` vs. `client_id` debacle, the solution that was decided upon was to stop generating *deprecated* + *experimental* attributes in codegen.

Effectively, these are attributes that not only never stabilized, but also will never stabilize.

Including those types will cause a conflict in codegen between two generated code names.  We only guarantee uniqueness of generated code for the set of (stable attributes and non-deprecated experimental attributes).

So this is actually updating Java codegen here.  I can continue generating the deprecated experimental attributes, but two things:

(1) You'll need to carry any namespace-related conflict resolution that causes in the java codegen going forward.
(2) We likely want some way to denote "This is deprecated for the Stable thing" vs. "This is deprecated and never use it again".

cc @lmolkova to correct me if I confused anything on this resolution.

buildscripts/checkstyle.xml
jsuereth: Great - I'd like to leave the changes as-is then.

buildscripts/templates/registry/incubating_java/IncubatingSemanticAttributes.java.j2
```diff
@@ -0,0 +1,66 @@
...
+    @Deprecated public static final {{ to_java_key_type(attribute) | trim }} {{ attribute.name | screaming_snake_case }} = {{ to_java_key_factory(attribute) | trim }}("{{ attribute.name }}");
+    {% endif %}
+    {% endfor %}
+    {% for attribute in ctx.attributes | sort(attribute="name") %}{% if attribute.type is mapping and attribute is experimental %}
+    // Enum definitions
+    /** Values for {@link #{{ attribute.name | screaming_snake_case }}}. */
+    public static final class {{ attribute.name | pascal_case }}Values {
+      {%- for member in attribute.type.members %}
+      {{ [member.brief or (member.id ~ '.')] | comment(indent=4) }}
+      public static final String {{ member.id | screaming_snake_case }} = "{{ member.value }}";
```
lmolkova, line 59: members could be integers - check out `rpc.grpc.status_code` (and we have `print_member_value` helper to write `0` vs `"string"`)

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/ContainerIncubatingAttributes.java
lmolkova: How to resolve client.id vs client_id - https://github.com/open-telemetry/semantic-conventions/issues/1031#issuecomment-2173887975 - manually drop `messaging.client_id`.

Existing jinja and build-tools solution implements it already and here's the weaver-based approach in python
- [declare a list of excluded ones](https://github.com/open-telemetry/opentelemetry-python/blob/889f7df41ecca926c7847f3c0913b7b9bbc0dd62/scripts/semconv/templates/registry/weaver.yaml#L7)
- [exclude them in jinja](https://github.com/open-telemetry/opentelemetry-python/blob/889f7df41ecca926c7847f3c0913b7b9bbc0dd62/scripts/semconv/templates/registry/semantic_attributes.j2#L22) or in JQ (in python we decided to generate them as comments for transparency) 


WRT if we should generate deprecated+experimental attributes - we should. The motivation - dependency version conflicts - there is always a risk that semconv artifact version is resolved to something you don't expect. 

If experimental+deprecated attributes are not generated, it could mean that two parts of app that use schema_url vN and schema_url N+1 can become incompatible and can't coexist since there is no semconv artifact that contains both.

By generating experiemental+deprecated we make it ~~much easier~~ possible to update different parts of user app independently

jsuereth: Ok, this is now fixed.

buildscripts/templates/registry/incubating_java/IncubatingSemanticAttributes.java.j2
jsuereth: It's funny - I fixed it in ONE template but not the other,thanks for catching this.

jsuereth: Fixed

buildscripts/templates/registry/incubating_java/weaver.yaml
jsuereth: These should be back now.

lquerel (commented):
Here are two minor suggestions:

1. Attributes are sorted by default, so most of the sort filters can be removed.
2. The file name definition can be simplified, as detailed below.

I recommend using the new, simpler method to specify file names. For example, in `IncubatingSemanticAttributes.java.j2`, instead of:

```jinja
{%- set my_file_name = ctx.root_namespace | pascal_case ~ "IncubatingAttributes.java" -%}
{{- template.set_file_name(my_file_name) -}}
```

you could define the file name directly in `weaver.yaml` within each template object like this:

```yaml
templates:
  - template: "incubating_java/IncubatingSemanticAttributes.java.j2"
    filter: ...
    application_mode: ...
    file_name: ctx.root_namespace | pascal_case ~ "IncubatingAttributes.java"
```

buildscripts/templates/registry/incubating_java/IncubatingSemanticAttributes.java.j2
```diff
@@ -0,0 +1,66 @@
...
+    {%- endif -%}
+    {{ [attribute.brief, concat_if("\n\nNotes:\n\n", attribute.note), deprecated_javadoc] | comment(indent=4) }}
+    {% if attribute is deprecated %}@Deprecated{% endif %} public static final {{ to_java_key_type(attribute) | trim }} {{ attribute.name | screaming_snake_case }} = {{ to_java_key_factory(attribute) | trim }}("{{ attribute.name }}");
+    {% elif attribute is stable %}
+    {%- set stable_class_link = "io.opentelemetry.semconv." ~ stable_class_name ~ "#" ~ (attribute.name | screaming_snake_case) -%}
+    {{ ["@deprecated deprecated in favor of stable {@link " ~ stable_class_link ~ "} attribute."] | comment(indent=4) }}
+    @Deprecated public static final {{ to_java_key_type(attribute) | trim }} {{ attribute.name | screaming_snake_case }} = {{ to_java_key_factory(attribute) | trim }}("{{ attribute.name }}");
+    {% endif %}
+    {% endfor %}
+    {% for attribute in ctx.attributes | sort(attribute="name") %}{% if attribute.type is mapping and attribute is experimental %}
```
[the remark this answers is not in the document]
jsuereth: fixed

jsuereth:
> Here are two minor suggestions:
> 
> 1. Attributes are sorted by default, so most of the sort filters can be removed.

Fixed

> 2. The file name definition can be simplified, as detailed below.
> 
> I recommend using the new, simpler method to specify file names. For example, in `IncubatingSemanticAttributes.java.j2`, instead of:
> 
> ```
> {%- set my_file_name = ctx.root_namespace | pascal_case ~ "IncubatingAttributes.java" -%}
> {{- template.set_file_name(my_file_name) -}}
> ```
> 
> you could define the file name directly in `weaver.yaml` within each template object like this:
> 
> ```yaml
> templates:
>   - template: "incubating_java/IncubatingSemanticAttributes.java.j2"
>     filter: ...
>     application_mode: ...
>     file_name: ctx.root_namespace | pascal_case ~ "IncubatingAttributes.java"
> ```

FYI - the `file_name` is actually a JINJA template, not an expression, so it's:

```
templates:
  - template: "incubating_java/IncubatingSemanticAttributes.java.j2"
    filter: ...
    application_mode: ...
    file_name: "{{ctx.root_namespace | pascal_case}}IncubatingAttributes.java"
```

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/MessagingIncubatingAttributes.java
```diff
@@ -343,15 +473,31 @@ public final class MessagingIncubatingAttributes {
...
   public static final AttributeKey<String> MESSAGING_SERVICEBUS_DISPOSITION_STATUS =
       stringKey("messaging.servicebus.disposition_status");
 
-  /** Number of deliveries that have been attempted for this message. */
+  /**
+   * Number of deliveries that have been attempted for this message.
+   *
+   * <p>
+   *
+   * @deprecated Replaced by {@code messaging.servicebus.destination.subscription_name}.
```
jack-berg, line 489: I don't think that this is actually deprecated. Applies to several other attributes in this file as well. https://github.com/open-telemetry/semantic-conventions/blob/main/docs/messaging/azure-messaging.md#span-attributes

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/HttpIncubatingAttributes.java
```diff
@@ -309,23 +277,18 @@ public static final class HttpConnectionStateValues {
     private HttpConnectionStateValues() {}
   }
 
-  /**
-   * Values for {@link #HTTP_FLAVOR}.
-   *
-   * @deprecated Deprecated, use `network.protocol.name` instead.
-   */
-  @Deprecated
```
jack-berg, line 317: Something's fishy. This is still deprecated.

https://github.com/open-telemetry/semantic-conventions/blob/5eb37f706894718493af7a668971c651acb80812/docs/attributes-registry/http.md#deprecated-http-attributes

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/HttpIncubatingAttributes.java
```diff
@@ -229,25 +195,15 @@ public final class HttpIncubatingAttributes {
...
-   *
-   * <p>Notes:
-   *
-   * <ul>
-   *   <li>MUST NOT be populated when this is not supported by the HTTP server framework as the
-   *       route attribute should have low-cardinality and the URI path can NOT substitute it.
-   *       SHOULD include the <a href="/docs/http/http-spans.md#http-server-definitions">application
-   *       root</a> if there is one.
-   * </ul>
-   *
```
jack-berg, line 250: Why does this description go away? Is it gone from the YAML model? Or it something with the template?

jsuereth: It's how I `if-else` the template. I can add them back if you wanted, but thesee are attributes that have moved from Incubating -> Stable so I didn't keep the description.

jsuereth: fixed

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/HttpIncubatingAttributes.java
jsuereth: I removed deprecated annotations from enum message definitions, adding them back in.

jsuereth: Fixed

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/ContainerIncubatingAttributes.java
```diff
@@ -123,19 +120,18 @@ public final class ContainerIncubatingAttributes {
 
   // Enum definitions
   /**
-   * Values for {@link #CONTAINER_CPU_STATE}.
```
lmolkova, line 126: oops

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/ErrorIncubatingAttributes.java
```diff
@@ -47,13 +54,11 @@ public final class ErrorIncubatingAttributes {
    * Values for {@link #ERROR_TYPE}.
    *
    * @deprecated deprecated in favor of stable {@link
-   *     io.opentelemetry.semconv.ErrorAttributes.ErrorTypeValues} attribute.
+   *     io.opentelemetry.semconv.ErrorAttributes#ERROR_TYPE} attribute.
```
lmolkova, line 57: nit: this used to point to the enum type and now points to the attribute name.

jsuereth: Fixed

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/ContainerIncubatingAttributes.java
jsuereth: Fixed

lmolkova (approved):

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/SystemIncubatingAttributes.java
```diff
@@ -260,11 +262,10 @@ private SystemProcessStatusValues() {}
   }
 
   /**
-   * Values for {@link #SYSTEM_PROCESSES_STATUS}.
+   * Values for {@link #SYSTEM_PROCESSES_STATUS}
    *
-   * @deprecated Deprecated, use `system.process.status` instead.
+   * @deprecated Replaced by {@code system.process.status}.
    */
-  @Deprecated
```
jack-berg, line 267: These `@Deprecated` annotations should not go away

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/ArtifactIncubatingAttributes.java
```diff
@@ -80,5 +75,7 @@ public final class ArtifactIncubatingAttributes {
   /** The version of the artifact. */
   public static final AttributeKey<String> ARTIFACT_VERSION = stringKey("artifact.version");
 
+  // Enum definitions
```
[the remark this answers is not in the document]
jack-berg: I don't think its a big deal. Your choice.

jack-berg:
@jsuereth, @lmolkova, @lquerel, and @open-telemetry/java-approvers - I'll merge this tomorrow if there are no additional comments.

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/CicdIncubatingAttributes.java
```diff
@@ -46,13 +45,13 @@ public final class CicdIncubatingAttributes {
...
   public static final class CicdPipelineTaskTypeValues {
-    /** build. */
+    /** build */
     public static final String BUILD = "build";
 
-    /** test. */
+    /** test */
     public static final String TEST = "test";
 
-    /** deploy. */
```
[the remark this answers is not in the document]
lquerel: The GH issue for tracking https://github.com/open-telemetry/weaver/issues/353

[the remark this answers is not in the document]
jsuereth: thanks @lquerel !

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/ArtifactIncubatingAttributes.java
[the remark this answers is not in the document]
jsuereth: Let's leave as-is for now.

Once we're on weaver (the important piece of this PR) there's incremental improvements we can begin making - but I'd like to focus on progress.

PR #73, opened 2024-06-03
Update to semantic-conventions 1.26.0

crossoverJie:
Update to semantic-conventions 1.26.0


semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/MessagingIncubatingAttributes.java
```diff
@@ -35,6 +35,14 @@ public final class MessagingIncubatingAttributes {
...
 
   /** A unique identifier for the client that consumes or produces a message. */
+  public static final AttributeKey<String> MESSAGING_CLIENT_ID = stringKey("messaging.client.id");
+
+  /**
+   * Deprecated, use {@code messaging.client.id} instead.
+   *
+   * @deprecated Deprecated, use `messaging.client.id` instead.
+   */
+  @Deprecated
```
crossoverJie, line 45: Due to duplicate variable names, CI cannot pass😂.

laurit: you could work around this by adding
```
  {# deprecated messaging.client_id clashes with messaging.client.id #}
  {%- if attribute.fqn != "messaging.client_id" %}
```
to https://github.com/open-telemetry/semantic-conventions-java/blob/756a10d3c3efcf7d87b2ac246e22d64cc394ddf6/buildscripts/templates/SemanticAttributes.java.j2#L67
@lmolkova is there a better solution for this?

trask: check out https://github.com/open-telemetry/semantic-conventions/issues/1031 and https://github.com/open-telemetry/semantic-conventions/issues/1118

we may want to go ahead and implement option 2 from https://github.com/open-telemetry/semantic-conventions/issues/1118 (but wait to merge it until that proposal is finalized)

I've added topic to today's Java SIG meeting to discuss

lmolkova: See https://github.com/open-telemetry/semantic-conventions-java/pull/75 - we'll provide build-tool fix ASAP as we have a concensus

lmolkova: Here's the change that drops problematic attribute - https://github.com/crossoverJie/semantic-conventions-java/pull/1 
(if Java SIG will accept it)

crossoverJie: > Here's the change that drops problematic attribute - [crossoverJie#1](https://github.com/crossoverJie/semantic-conventions-java/pull/1) (if Java SIG will accept it as a mitigation)

Thank you for your work on this issue.

trask (approved):

jack-berg (approved):

PR #76, opened 2024-06-09
Collisions: manual conflict resolution

lmolkova:
Alternative to https://github.com/open-telemetry/semantic-conventions-java/pull/75:
- `_` and `.` are rendered the same in the code
- semconv policy does not allow to have non-deprecated attributes with the same const name. 
- semconv policy allows to have a collision with a deprecated attribute
- there is a manual conflict resolution way that allows to change constant names for arbitrary attributes and/or drop them


semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/MessagingIncubatingAttributes.java
```diff
@@ -35,7 +35,7 @@ public final class MessagingIncubatingAttributes {
       longKey("messaging.batch.message_count");
 
   /** A unique identifier for the client that consumes or produces a message. */
-  public static final AttributeKey<String> MESSAGING_CLIENT_ID = stringKey("messaging.client_id");
+  public static final AttributeKey<String> MESSAGING_CLIENT_ID = stringKey("messaging.client.id");
```
lmolkova, line 38: that's the demo of the resolution:
- `messaging.client_id` is deprecated, `messaging.client.id` is not 
- `messaging.client_id` is completely ignored during codegen

This results in potential mismatch between schema url and attribute values, but we would not rename a stable attribute, but it seems tolerable for experimental.

buildscripts/templates/SemanticAttributes.java.j2
```diff
@@ -39,6 +39,10 @@
 {%- set filtered_enums = filtered_attributes | selectattr('is_enum', 'equalto', true) | list %}
 {%- set excluded_namespaces_list = excluded_namespaces.replace("\"", "").split(' ') %}
 {%- if root_namespace not in excluded_namespaces_list and filtered_attributes | count > 0 %}
+{%- set name_map ={"messaging.client.id":"messaging_client_id2"} -%}
```
lmolkova, line 42: manually defined and maintained maintained list of name overwrites. 

TODO:
- [ ] it should be easier and done with `-D`
- [ ]  it should be used in class names
- [ ] it should not affect stable artifact
- [ ] javadoc on incubating `ID2` should refer to stable `ID`

semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/MessagingIncubatingAttributes.java
lmolkova: no relevant anymore

build.gradle.kts
```diff
@@ -86,26 +86,32 @@ fun generateTask(taskName: String, incubating: Boolean) {
     val outputDir = if (incubating) "semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/" else "semconv/src/main/java/io/opentelemetry/semconv/"
     val packageNameArg = if (incubating) "io.opentelemetry.semconv.incubating" else "io.opentelemetry.semconv"
     val stablePackageNameArg = if (incubating) "io.opentelemetry.semconv" else ""
+    val excludedNamespaces = """[\"ios\",\"aspnetcore\",\"signalr\"]"""
+    val nameMapping = """{\"http.request.method\":\"http.request.method_new\"}"""
+    val excludedAttributes = """[\"http.flavor\"]"""
```
trask, line 91: 👍

PR #83, opened 2024-08-19
Include Incubating prefix on enum inner class names

### LINKED ISSUE #67: Add "Incubating" to nested class names?
wdyt of renaming `DbSystemValues` to `DbIncubatingSystemValues` to make usage clearer when importing the nested class directly (and also possibly also since individual enum attribute values can be declared stable / experimental and so it may be more common to need both `DbSystemValues` to `DbIncubatingSystemValues`)

cc @lmolkova @jack-berg

jack-berg:
Resolves #67.


breedx-splk (approved):
Well, it sure is explicit. 🙃

trask (approved):

PR #87, opened 2024-08-23
add OSGi manifest headers for jars

### LINKED ISSUE #86: Add OSGi manifest headers
The current build of the semantic-conventions-java project lacks the necessary OSGi manifest headers in its generated JAR files. This omission prevents the JARs from being used directly as OSGi bundles. To ensure compatibility and ease of integration with OSGi environments, it is important to include the appropriate OSGi headers in the manifest files of these JARs.

Adding these headers would significantly enhance the usability of this library in OSGi-based projects, allowing for seamless integration into environments that rely on modularity and dynamic loading of components. It would also align this project with best practices for Java library distribution in modular systems.

**Proposed Solution**

Modify the build configuration to automatically include the necessary OSGi headers in the manifest files of all JARs produced by this project.

bdhoine:
resolves #86


jack-berg:
There's a similar PR in `opentelemetry-java` right now: https://github.com/open-telemetry/opentelemetry-java/pull/6669

My comment there is applicable here as well: 

> As someone without OSGi experience, I think its important to add some sort of end-to-end integration test. A sample app built with OSGi which consumes these artifacts and demonstrates it works as intended.

bdhoine:
@jack-berg yes I noticed, @royteeuwen and myself are working on some details how to solve the current issues we are seeing while testing, this topic definitely needs some more work!

PR #130, opened 2025-01-06
Prepare stable release

### LINKED ISSUE #90: Stabilize opentelemetry-semconv
Want to start a conversation about what stands in the way of marking the `opentelemetry-semconv` artifact as stable. We've come a long way recently, re-organizing by domain, switching from build-tools to weaver based code generation, and changing how we think about deprecated / incubating attributes.

jack-berg:
Resolves #90.

- Remove -alpha suffix from `opentelemetry-semconv` artifact
- Remove deprecated ResourceAttributes, SemanticAttributes
- Update japicmp to generate diff for first stable release


docs/apidiffs/current_vs_latest/opentelemetry-semconv.txt
```diff
@@ -0,0 +1,170 @@
...
+	+++  NEW FIELD: PUBLIC(+) STATIC(+) FINAL(+) io.opentelemetry.api.common.AttributeKey<java.lang.String> ERROR_TYPE
++++  NEW CLASS: PUBLIC(+) STATIC(+) FINAL(+) io.opentelemetry.semconv.ErrorAttributes$ErrorTypeValues  (not serializable)
+	+++  CLASS FILE FORMAT VERSION: 52.0 <- n.a.
+	+++  NEW SUPERCLASS: java.lang.Object
+	+++  NEW FIELD: PUBLIC(+) STATIC(+) FINAL(+) java.lang.String OTHER
++++  NEW CLASS: PUBLIC(+) FINAL(+) io.opentelemetry.semconv.ExceptionAttributes  (not serializable)
+	+++  CLASS FILE FORMAT VERSION: 52.0 <- n.a.
+	+++  NEW SUPERCLASS: java.lang.Object
+	+++  NEW FIELD: PUBLIC(+) STATIC(+) FINAL(+) io.opentelemetry.api.common.AttributeKey<java.lang.String> EXCEPTION_TYPE
+	+++  NEW FIELD: PUBLIC(+) STATIC(+) FINAL(+) io.opentelemetry.api.common.AttributeKey<java.lang.Boolean> EXCEPTION_ESCAPED
```
trask, line 32: should we proactively remove this unused (and likely to be deprecated) constant? https://github.com/open-telemetry/semantic-conventions/issues/1516

cc @lmolkova

docs/apidiffs/current_vs_latest/opentelemetry-semconv.txt
```diff
@@ -0,0 +1,170 @@
...
+	+++  CLASS FILE FORMAT VERSION: 52.0 <- n.a.
+	+++  NEW SUPERCLASS: java.lang.Object
+	+++  NEW FIELD: PUBLIC(+) STATIC(+) FINAL(+) java.lang.String V1_24_0
+	+++  NEW FIELD: PUBLIC(+) STATIC(+) FINAL(+) java.lang.String V1_23_1
+	+++  NEW FIELD: PUBLIC(+) STATIC(+) FINAL(+) java.lang.String V1_22_0
+	+++  NEW FIELD: PUBLIC(+) STATIC(+) FINAL(+) java.lang.String V1_29_0
+	+++  NEW FIELD: PUBLIC(+) STATIC(+) FINAL(+) java.lang.String V1_28_0
+	+++  NEW FIELD: PUBLIC(+) STATIC(+) FINAL(+) java.lang.String V1_27_0
+	+++  NEW FIELD: PUBLIC(+) STATIC(+) FINAL(+) java.lang.String V1_26_0
+	+++  NEW FIELD: PUBLIC(+) STATIC(+) FINAL(+) java.lang.String V1_25_0
```
trask, line 127: any thoughts on this list (i.e. how far to go back)?

cc @lmolkova

trask (commented):
👍

docs/apidiffs/current_vs_latest/opentelemetry-semconv.txt
jack-berg: > should we proactively remove this unused (and likely to be deprecated) constant?

Would you imagine doing this in `semantic-conventions` or in the code generation tooling or in the code generation configuration local to `semantic-convention-java`?

docs/apidiffs/current_vs_latest/opentelemetry-semconv.txt
jack-berg: Notable versions:

- 1.21.0 was first release from `semantic-conventions` repo
- 1.23.0 stable http conventions
- 1.24.0 stable JVM conventions, significant restructuring of `semantic-convention-java` to switch to `{Domain}Attributes` naming convention and breakout `opentelemetry-semconv` and `opentelemetry-semconv-incubating` artifacts. 
- 1.28.0 switch code gen for `semantic-convention-java` to weaver

We could start conservative with the schema URLs we include and go further back later if needed?

trask: yeah, I don't really have any specific concerns or objection with the existing list

> 1.21.0 was first release from `semantic-conventions` repo

this seems as good as any reason, but agree we can add later, only thing that would affect this PR / release would be if we wanted to restrict the list

> 1.23.0 stable http conventions

btw, I noticed `1.23.0` is missing in the list (but `1.22.0` and `1.23.1` are included)

docs/apidiffs/current_vs_latest/opentelemetry-semconv.txt
trask: I think we would need to hack it somehow to remove it in this repo, and wait to see what happens to the attribute in semconv

not sure it's worth it, there will be other deprecated "stable" attributes in the future anyways...

lmolkova: It should be really easy to exclude (a few) specific attributes by extending this list -

https://github.com/open-telemetry/semantic-conventions-java/blob/828296fad46df5651ba0d83fd8e52f09a0a14c9b/buildscripts/templates/registry/java/weaver.yaml#L3

It would felt great if we already deprecated it in semconv.

trask: thanks, I'll send a separate PR to try it out

docs/apidiffs/current_vs_latest/opentelemetry-semconv.txt
lmolkova: I did a quick search for [`"SchemaUrls.V1_" NOT is:fork`](https://github.com/search?q=%22SchemaUrls.V1_%22+NOT+is%3Afork&type=code) across GitHub and it seems there are just a few places that use it.

I only see usages of V1_24_0 and above - which makes sense - we ask to stay at 1.24.0 in DB and messaging migration plan.

I'd remove anything below 1.24.0.

docs/apidiffs/current_vs_latest/opentelemetry-semconv.txt
jack-berg: Just so we're on the same page, the principle behind removing this attribute is that we want to start with a clean slate and not stabilize anything deprecated unnecessarily. By this logic, this is the only instance where we would do this, since any attribute which is deprecated in the future will have to be maintained for backwards compatibility.

Is this your understanding?

trask: yes

docs/apidiffs/current_vs_latest/opentelemetry-semconv.txt
trask: ok, I'll send a PR to remove 1.22.0 and 1.23.1

docs/apidiffs/current_vs_latest/opentelemetry-semconv.txt
jack-berg: Makes sense to me. If for some strange reason we need this attribute, we can revert the exclusion to restore it.

trask: #133

docs/apidiffs/current_vs_latest/opentelemetry-semconv.txt
trask: #134

lmolkova: Quick thought - would we regret having exact patch in the version?
Should we do `V1_25` (`V1_25_X`) instead similarly to release branches?

I.e. if semconv is patched, we'd not add a new schema version.  
Happy to send a PR if you folks think it makes sense.

trask: I kind of agree, but if we're saying there's no difference between patched versions, shouldn't the schema url itself be `https://opentelemetry.io/schemas/1.29` or `https://opentelemetry.io/schemas/1.29.x`?

jack-berg: That's a can of worms though because these are resolvable URLs, and `../1.29` and `../1.29.x` do not resolve. I think our schema URLs should reflect whatever the publishing convention is on opentelemetry.io. We could open an issue about this and block until resolved, but I don't think we should block stability. If opentelemetry.io later updates to reflect one of the conventions you propose, we can update our schema URL values from that point onward.

lmolkova: ok, I created https://github.com/open-telemetry/opentelemetry-specification/issues/4355 - we can change the format of the constant later if it comes through.

trask:
sorry, last question, is this preparing an RC release, or preparing a stable release?

jack-berg:
> sorry, last question, is this preparing an RC release, or preparing a stable release?

I was thinking that we execute an RC manually, locally. Haven't thought about what (temporary?) changes would be needed to cut an RC. If I did the release locally, I was thinking I would open a branch to communicate the changes I'd be making.

trask (approved):

lmolkova (approved):
🚀

gabrieljones:
> Remove deprecated ResourceAttributes, SemanticAttributes
Will this be a 2.x release or is this breaking change going to occur without bumping the major version number?

gabrieljones:
> Remove deprecated ResourceAttributes, SemanticAttributes
@jack-berg 

Will this be a 2.x release or is this breaking change going to occur without bumping the major version number?

jack-berg:
> Will this be a 2.x release or is this breaking change going to occur without bumping the major version number?

This is a breaking change without bumping the major version number:

- Breaking changes are allowed (though still taken seriously) since the artifact has the `-alpha` suffix. After a stable release, no breaking changes will be allowed without a major version bump.
- This repo is versioned in lockstep with [semantic-conventions](https://github.com/open-telemetry/semantic-conventions), so `v1.30.0` here mirrors `semantic-conventions` release `v1.30.0`. We've worked carefully with the `semantic-conventions` maintainers to ensure that they can't make any changes in a minor version that would necessitate a breaking change in the generated code here.

PR #132, opened 2025-01-08
Update renovate config

trask:


zeitlinger:
seems like there should be a "renovate why" to see what is done for a particular dependency

trask:
> seems like there should be a "renovate why" to see what is done for a particular dependency

have you seen the link to the renovate log at the bottom of the PR description on all renovate PRs?

it still took me some time to figure out the problem from the log, but it's pretty helpful

zeitlinger:
> it still took me some time to figure out the problem from the log, but it's pretty helpful

been there - it's possible, but I ended up reading some renovate source code to actually figure out why something was not updated

laurit (approved):

PR #133, opened 2025-01-08
Exclude exception.escaped from (to-be-stable) artifact

trask:
See https://github.com/open-telemetry/semantic-conventions-java/pull/130#discussion_r1904658679

cc @lmolkova

### DIFF
--- a/buildscripts/templates/registry/incubating_java/weaver.yaml
+++ b/buildscripts/templates/registry/incubating_java/weaver.yaml
@@ -1,6 +1,6 @@
 params:
   excluded_namespaces: ["ios", "aspnetcore", "signalr"]
-  excluded_attributes: ["messaging.client_id"]
+  excluded_attributes: ["messaging.client_id", "exception.escaped"]
 comment_formats:
   javadoc:
     format: html
--- a/buildscripts/templates/registry/incubating_java/weaver.yaml
+++ b/buildscripts/templates/registry/incubating_java/weaver.yaml
@@ -19,12 +19,11 @@ templates:
     filter: > 
       semconv_grouped_attributes({
         "exclude_root_namespace": $excluded_namespaces,
-        "exclude_stability": [],
-        "exclude_deprecated": false
+        "exclude_stability": []
       }) | map({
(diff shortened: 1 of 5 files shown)


buildscripts/templates/registry/incubating_java/weaver.yaml
```diff
@@ -19,12 +19,11 @@ templates:
     filter: > 
       semconv_grouped_attributes({
         "exclude_root_namespace": $excluded_namespaces,
-        "exclude_stability": [],
-        "exclude_deprecated": false
```
trask, line 23: exclude_deprecated didn't seem to be needed

buildscripts/templates/registry/incubating_java/weaver.yaml
```diff
@@ -19,12 +19,11 @@ templates:
...
+        "exclude_stability": []
       }) | map({
           root_namespace: .root_namespace,
           attributes: .attributes,
-          excluded_attributes: $excluded_attributes[]
+          excluded_attributes: $excluded_attributes
```
trask, line 26: without this it was only working with a single item in the excluded_attributes list

trask:
oops, need to fix javadoc, incubating constant is deprecated referencing the stable constant

jack-berg (approved):

lmolkova:
> oops, need to fix javadoc, incubating constant is deprecated referencing the stable constant

maybe we should exclude it from incubating artifact too?

trask:
> > oops, need to fix javadoc, incubating constant is deprecated referencing the stable constant
> 
> maybe we should exclude it from incubating artifact too?

yeah, this is certainly easier, we can always bring it back (as deprecated) in the incubating artifact later

trask:
> > > oops, need to fix javadoc, incubating constant is deprecated referencing the stable constant
> > 
> > 
> > maybe we should exclude it from incubating artifact too?
> 
> yeah, this is certainly easier, we can always bring it back (as deprecated) in the incubating artifact later

@jack-berg I reset your approval to make sure you're ok with this

lmolkova (approved):

PR #134, opened 2025-01-08
Remove old schema urls before stabilizing

trask:
cc @lmolkova

### DIFF
--- a/semconv/src/main/java/io/opentelemetry/semconv/SchemaUrls.java
+++ b/semconv/src/main/java/io/opentelemetry/semconv/SchemaUrls.java
@@ -13,8 +13,6 @@ public final class SchemaUrls {
   public static final String V1_26_0 = "https://opentelemetry.io/schemas/1.26.0";
   public static final String V1_25_0 = "https://opentelemetry.io/schemas/1.25.0";
   public static final String V1_24_0 = "https://opentelemetry.io/schemas/1.24.0";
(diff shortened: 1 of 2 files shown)


semconv/src/main/java/io/opentelemetry/semconv/SchemaUrls.java
```diff
@@ -13,8 +13,6 @@ public final class SchemaUrls {
   public static final String V1_26_0 = "https://opentelemetry.io/schemas/1.26.0";
   public static final String V1_25_0 = "https://opentelemetry.io/schemas/1.25.0";
   public static final String V1_24_0 = "https://opentelemetry.io/schemas/1.24.0";
-  public static final String V1_23_1 = "https://opentelemetry.io/schemas/1.23.1";
-  public static final String V1_22_0 = "https://opentelemetry.io/schemas/1.22.0";
```
lmolkova, line 17: I think you have to remove them from https://github.com/open-telemetry/semantic-conventions-java/blob/828296fad46df5651ba0d83fd8e52f09a0a14c9b/build.gradle.kts#L22 too

lmolkova (approved):

jack-berg (approved):

jack-berg:
Linking the conversation that led to this: https://github.com/open-telemetry/semantic-conventions-java/pull/130#discussion_r1904661652

PR #141, opened 2025-01-27
Bump to semantic-conventions v1.30.0

jack-berg:


buildscripts/templates/registry/incubating_java/weaver.yaml
```diff
@@ -1,5 +1,5 @@
 params:
-  excluded_namespaces: ["ios", "aspnetcore", "signalr"]
+  excluded_namespaces: ["dotnet", "ios", "aspnetcore", "signalr"]
```
jack-berg, line 2: Heads up - excluding `dotnet.` attributes as a continuation of https://github.com/open-telemetry/semantic-conventions-java/issues/48

trask (approved):

PR #151, opened 2025-01-29
Fix gradle warnings

jack-berg:
Fixes: 
```
w: file:///Users/jberg/code/open-telemetry/semantic-conventions-java/build.gradle.kts:69:10: 'getter for buildDir: File!' is deprecated. Deprecated in Java
w: file:///Users/jberg/code/open-telemetry/semantic-conventions-java/build.gradle.kts:82:10: 'getter for buildDir: File!' is deprecated. Deprecated in Java
w: file:///Users/jberg/code/open-telemetry/semantic-conventions-java/build.gradle.kts:108:39: 'getter for buildDir: File!' is deprecated. Deprecated in Java
'indentWithSpaces' is deprecated, use 'leadingTabsToSpaces' in your gradle build script instead.
'indentWithSpaces' is deprecated, use 'leadingTabsToSpaces' in your gradle build script instead.

```

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -79,7 +79,7 @@ val unzipConfigurationSchema by tasks.registering(Copy::class) {
     val pathParts = path.split("/")
     path = pathParts.subList(1, pathParts.size).joinToString("/")
   })
-  into(layout.buildDirectory.file("semantic-conventions-${semanticConventionsVersion}/"))
+  into(layout.buildDirectory.dir("semantic-conventions-${semanticConventionsVersion}/"))
 }
 
 fun generateTask(taskName: String, incubating: Boolean) {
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -66,7 +66,7 @@ val schemaUrl = "https://opentelemetry.io/schemas/$semanticConventionsVersion"
 
 val downloadSemanticConventions by tasks.registering(Download::class) {
   src(semanticConventionsRepoZip)
-  dest("$buildDir/semantic-conventions-${semanticConventionsVersion}/semantic-conventions.zip")
+  dest("${layout.buildDirectory.asFile.get()}/semantic-conventions-${semanticConventionsVersion}/semantic-conventions.zip")
```
laurit, line 69: from what I understood the intended usage would be `dest(layout.buildDirectory.file("semantic-conventions-${semanticConventionsVersion}/semantic-conventions.zip"))` (didn't test) to allow for lazy resolution of the build dir

laurit (approved):

build.gradle.kts
jack-berg: Will take a look. Thanks for the advice!

jack-berg: Had to drop the leading `/` but works otherwise

build.gradle.kts
```diff
@@ -79,7 +79,7 @@ val unzipConfigurationSchema by tasks.registering(Copy::class) {
     val pathParts = path.split("/")
     path = pathParts.subList(1, pathParts.size).joinToString("/")
   })
-  into("$buildDir/semantic-conventions-${semanticConventionsVersion}/")
+  into(layout.buildDirectory.file("semantic-conventions-${semanticConventionsVersion}/"))
```
laurit, line 82: I think this should use `dir` instead of `file`. Idk whether using `file` here breaks anything but the api seems to distinguish between regular files and directories.

PR #155, opened 2025-02-07
Update renovate config to pin github action versions

trask:


breedx-splk (approved):

breedx-splk:
Does renovate then PR updates that are also pinned then?

trask:
> Does renovate then PR updates that are also pinned then?

yup! (and it updates the comment that contains the version number which is super nice since the hashes are not human readable)

Issue #208, opened 2025-04-23
Don't generate attribute keys with unknown types


lmolkova:
Semconv tooling is being expanded to support attributes with `any` value (on logs/events for the time being) in https://github.com/open-telemetry/weaver/pull/707.

Once such attribute is defined in semconv (ETA 1-2 month from now), Java code generation here

https://github.com/open-telemetry/semantic-conventions-java/blob/3c7edf34f5b548e64a63793e3772da40754d874c/buildscripts/templates/registry/incubating_java/IncubatingSemanticAttributes.java.j2#L44

will start producing invalid code since there is no mapping defined for `any` in [java_template_key_factory](https://github.com/open-telemetry/semantic-conventions-java/blob/3c7edf34f5b548e64a63793e3772da40754d874c/buildscripts/templates/registry/incubating_java/weaver.yaml#L58) and it's not trivial to add it (given incubating `ExtendedAttributes` and other uncertainties around this).

The recommendation is to update jinja to exclude all `any` (and other future unknown) attribute types from codegen for the time being.

Note: another supported variation of this would be `template[any]`

PR #220, opened 2025-05-15
Handle mix of stable and unstable enum values

trask:
Fixing https://github.com/open-telemetry/semantic-conventions-java/pull/217#discussion_r2085472786 before upgrading to semconv 1.33.0.

### DIFF
--- a/semconv/src/main/java/io/opentelemetry/semconv/NetworkAttributes.java
+++ b/semconv/src/main/java/io/opentelemetry/semconv/NetworkAttributes.java
@@ -78,6 +78,7 @@ public final class NetworkAttributes {
   // Enum definition
   /** Values for {@link #NETWORK_TRANSPORT}. */
   public static final class NetworkTransportValues {
+
     /** TCP */
     public static final String TCP = "tcp";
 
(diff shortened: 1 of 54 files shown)


semconv/src/main/java/io/opentelemetry/semconv/NetworkAttributes.java
```diff
@@ -90,15 +91,13 @@ public static final class NetworkTransportValues {
     /** Unix domain socket */
     public static final String UNIX = "unix";
 
-    /** QUIC */
-    public static final String QUIC = "quic";
-
```
trask, line 95: oops, this has been around for a while 🙈

@jack-berg any thoughts? I'm guessing we need to wedge it back into the stable definition here somehow?

trentm: https://github.com/open-telemetry/opentelemetry-js/pull/5690#issuecomment-2885209525 is the equivalent open question for the JS semconv package

trask: see https://github.com/open-telemetry/semantic-conventions/pull/2275!

trask:
Closing, need to merge this into #208 in order to pick up the stability change to `QUIC`

PR #237, opened 2025-06-09
Pin weaver sha

trask:
Dependabot runs keep failing while trying to do this, see #103.

Hopefully it will be ok after this initial pinning.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -112,7 +112,7 @@ fun generateTask(taskName: String, incubating: Boolean) {
         "--mount", "type=bind,source=${modelPath},target=/home/weaver/source,readonly",
         "--mount", "type=bind,source=$projectDir/buildscripts/templates,target=/home/weaver/templates,readonly",
         "--mount", "type=bind,source=$projectDir/$outputDir,target=/home/weaver/target",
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -112,7 +112,7 @@ fun generateTask(taskName: String, incubating: Boolean) {
         "--mount", "type=bind,source=${modelPath},target=/home/weaver/source,readonly",
         "--mount", "type=bind,source=$projectDir/buildscripts/templates,target=/home/weaver/templates,readonly",
         "--mount", "type=bind,source=$projectDir/$outputDir,target=/home/weaver/target",
-        "otel/weaver:v0.15.2",
+        "otel/weaver:v0.15.2@sha256:b13acea09f721774daba36344861f689ac4bb8d6ecd94c4600b4d590c8fb34b9",
```
trask, line 115: hm, I don't think this is meaningful from docker perspective (it works, but it also works whatever text I put before the `@sha`)

trask: Renovate supports this as a way to "document" the version number, see https://github.com/open-telemetry/opentelemetry-java-examples/commit/763a32870a47a2dd6ab4830746c058bc9cbcee66

PR #294, opened 2025-09-06
Update link checking

trask:


trask:
@zeitlinger bump

zeitlinger (approved):
ok, let's do the rest when lychee is bumped

trask:
> ok, let's do the rest when lychee is bumped

I bumped it now 👍

PR #317, opened 2025-10-03
Generate constants for metric definitions

trask:
Related to https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/9478

Another option is to provide an API that creates the metric using these constants internally, though that doesn't help us in library instrumentation for all of the unstable metrics at least where we don't want to have a dependency on unstable semconv artifact, but we still want the constants for verification in tests.


trask:
> I think we should only generate incubating constants to start (but include stable metrics in the incubating files) to start.

will do

PR #328, opened 2025-10-17
Refactor CI build

trask:
Following other Java repos better


.github/workflows/build.yml
```diff
@@ -4,115 +4,12 @@ on:
   push:
     branches:
       - main
-  pull_request:
+      - release/*
```
trask: build wasn't running on push to release branch previously

PR #359, opened 2025-11-25
Modernize/simplify JUnit Gradle configuration

trask:
Use useJUnitJupiter() which automatically adds the necessary JUnit dependencies, removing the need to explicitly declare junit-bom, junit-jupiter-api, junit-jupiter-params, junit-jupiter-engine, and junit-platform-launcher.

### DIFF
--- a/buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
@@ -131,15 +131,12 @@ dependencies {
 
 testing {
   suites.withType(JvmTestSuite::class).configureEach {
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
```diff
@@ -131,15 +131,12 @@ dependencies {
 
 testing {
   suites.withType(JvmTestSuite::class).configureEach {
+    useJUnitJupiter("5.14.0")
```
trask: intentionally 1 version behind to confirm renovate picks it up

PR #463, opened 2026-04-29
Fix auto-update semconv workflow

trask:
https://github.com/open-telemetry/semantic-conventions-java/actions/runs/25111012721/job/73585353811


semconv/src/main/java/io/opentelemetry/semconv/SchemaUrls.java
```diff
@@ -6,7 +6,6 @@
 package io.opentelemetry.semconv;
 
 public final class SchemaUrls {
-
```
trask, line 9: this fixes https://github.com/open-telemetry/semantic-conventions-java/pull/461/changes#r3158951137

laurit (approved):

PR #489, opened 2026-05-26
introduce event semantic conventions

breedx-splk:
This introduces weaver config and templates for generating classes that contain event name constants.

This ultimately allows us to reference these constants from java/kotlin/android code without having to duplicate the names, which can be error prone and lead to inconsistencies.


breedx-splk:
Bump?

jack-berg (approved):

jack-berg:
FYI, I've pushed [4d55cf8](https://github.com/open-telemetry/semantic-conventions-java/pull/489/commits/4d55cf897ebdc7486780ecf35be32e989ae3b2d3) and [c8b90b3](https://github.com/open-telemetry/semantic-conventions-java/pull/489/commits/c8b90b339c535a350112860ce636a96ebc7c8eca). These prepared event constant generation for publishing stable constants, and brought the generation into alignment with attribute generation (things like javadoc formatting, where filtering takes place, use of @Deprecation annotation). Then, I commented out the bits for stable constant generation.

Once we're ready to publish constants for stable conventions, we simply uncomment those bits.

jaydeluca (approved):

PR #498, opened 2026-06-14
Add native OSGi bundle metadata to published jars

### LINKED ISSUE #494: OSGi support
With https://github.com/open-telemetry/opentelemetry-java/pull/8417, core modules support OSGi.

Would be good if other modules published within the opentelemetry java ecosystem did as well.

royteeuwen:
## Description

Closes #494

Publishes the `semconv` and `semconv-incubating` jars with OSGi bundle metadata so they can be consumed directly in OSGi containers, without external re-wrapping (e.g. hand-maintained wrapper bundles).

This mirrors the OSGi support recently added to [opentelemetry-java](https://github.com/open-telemetry/opentelemetry-java): the [bnd Gradle plugin](https://github.com/bndtools/bnd) (`biz.aQute.bnd.builder`) is wired into the shared `otel.java-conventions` plugin. Since the semconv modules have no SPI / ServiceLoader needs, the configuration is a trimmed-down version of the upstream one (no ServiceLoader mediator capabilities). Both published modules want OSGi metadata, so the bundle config is applied unconditionally rather than behind a flag.

## Changes

- `buildSrc/build.gradle.kts` — add the `biz.aQute.bnd:biz.aQute.bnd.gradle` plugin dependency.
- `otel.java-conventions.gradle.kts` — apply `biz.aQute.bnd.builder`, make archives reproducible, and configure the main `jar` task's bundle with `-exportcontents: io.opentelemetry.*`. `Import-Package` is left to bnd's auto-detection, except `io.opentelemetry.api.*` which is pinned to `[1.33,2)` (bnd can't infer a range there because the pinned compileOnly `opentelemetry-api` 1.33.0 isn't itself a bnd bundle; `1.33` is the baseline `:dependencyManagement` already requires). Both modules (and any future module) are bundled automatically; the `-sources`/`-javadoc` jars are left plain.
- `osgi-test/` — new `:osgi-test` module (not published) that validates the bundles in a real OSGi runtime (see Testing).

## Generated manifest (excerpt)

`opentelemetry-semconv`:
```
Bundle-SymbolicName: opentelemetry-semconv
Export-Package: io.opentelemetry.semconv;version="1.43.0";uses:="io.opentelemetry.api.common"
Import-Package: io.opentelemetry.api.common;version="[1.33,2)",java.lang,...
Automatic-Module-Name: io.opentelemetry.semconv
```

`opentelemetry-semconv-incubating`:
```
Bundle-SymbolicName: opentelemetry-semconv-incubating
Export-Package: io.opentelemetry.semconv.incubating;version="1.43.0";uses:="io.opentelemetry.api.common,io.opentelemetry.semconv"
Import-Package: io.opentelemetry.api.common;version="[1.33,2)",io.opentelemetry.semconv;version="[1.43,2)",...
```

## Testing

- New `:osgi-test` module boots the `semconv` + `semconv-incubating` jars inside a real Apache Felix container and runs JUnit 5 inside it, via bnd's `Bundle` → `Resolve` → `TestOSGi` task chain (modeled on opentelemetry-java's `integration-tests/osgi`). If either bundle's generated OSGi metadata is wrong (bad exports / unsatisfiable imports) the bundle won't resolve and the tests won't run — and the task fails explicitly in that case, since bnd otherwise reports success on zero tests. It's wired into `test`/`check`, so it runs as part of the normal `./gradlew build`.
- `./gradlew build` passes (unit tests, checkstyle, japicmp, animal-sniffer, and the OSGi resolve test).
- Verified the main jars carry the OSGi headers and the `-sources`/`-javadoc` jars remain plain.

### DIFF
--- a/buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
@@ -8,10 +8,13 @@ plugins {
   eclipse
   idea
 
+  id("biz.aQute.bnd.builder")
   id("otel.spotless-conventions")
 }
 
 val otelJava = extensions.create<OtelJavaExtension>("otelJava")
+otelJava.osgiEnabled.convention(true)
+otelJava.osgiOptionalPackages.convention(emptyList())
 
 java {
   toolchain {
@@ -27,6 +30,13 @@ checkstyle {
   toolVersion = "13.5.0"
   isIgnoreFailures = false
   configProperties["rootDir"] = rootDir
+}
+
+// normalize timestamps and file ordering in jars, making the outputs (including OSGi
+// manifests) reproducible. see open-telemetry/opentelemetry-java#4488
+tasks.withType<AbstractArchiveTask>().configureEach {
+  isPreserveFileTimestamps = false
+  isReproducibleFileOrder = true
 }
 
 val testJavaVersion = gradle.startParameter.projectProperties.get("testJavaVersion")?.let(JavaVersion::toVersion)
@@ -82,6 +92,30 @@ tasks {
       addBooleanOption("Xdoclint:all,-missing", true)
       // non-standard option to fail on warnings, see https://bugs.openjdk.java.net/browse/JDK-8200363
       addStringOption("Xwerror", "-quiet")
+    }
+  }
+
+  afterEvaluate {
+    if (otelJava.osgiEnabled.get()) {
+      named<Jar>("jar") {
+        // Configure OSGi metadata. semconv has no SPI / ServiceLoader needs, so this is the
+        // trimmed form of opentelemetry-java's otel.java-conventions OSGi config.
+        bundle {
+          // Modules may declare optional imports (typically compileOnly deps). The trailing
+          // "*" lets BND auto-import everything else (e.g. io.opentelemetry.api.*).
+          val optionalPackages = otelJava.osgiOptionalPackages.get()
+          val optionalImports =
+            optionalPackages.joinToString(",") { "$it.*;resolution:=optional;version=\"\${@}\"" }
+          val importPackages = if (optionalImports.isEmpty()) "*" else "$optionalImports,*"
+
+          bnd(
+            mapOf(
+              "-exportcontents" to "io.opentelemetry.*",
+              "Import-Package" to importPackages,
+            ),
+          )
+        }
+      }
     }
   }
 
--- a/buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
@@ -8,10 +8,13 @@ plugins {
   eclipse
   idea
 
+  id("biz.aQute.bnd.builder")
   id("otel.spotless-conventions")
 }
 
 val otelJava = extensions.create<OtelJavaExtension>("otelJava")
+otelJava.osgiEnabled.convention(true)
+otelJava.osgiOptionalPackages.convention(emptyList())
 
 java {
   toolchain {
@@ -27,6 +30,13 @@ checkstyle {
   toolVersion = "13.8.0"
   isIgnoreFailures = false
   configProperties["rootDir"] = rootDir
+}
+
+// normalize timestamps and file ordering in jars, making the outputs (including OSGi
+// manifests) reproducible. see open-telemetry/opentelemetry-java#4488
+tasks.withType<AbstractArchiveTask>().configureEach {
+  isPreserveFileTimestamps = false
+  isReproducibleFileOrder = true
 }
 
 val testJavaVersion = gradle.startParameter.projectProperties.get("testJavaVersion")?.let(JavaVersion::toVersion)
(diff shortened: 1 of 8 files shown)


royteeuwen:
@trask  anything I can do to get this mergeable?

trask (commented):
can you add a test?

or, related to https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/18995, is there maybe another way / tool we could validate a jar for osgi compliance?

royteeuwen:
@trask added a test. I went with the same approach as opentelemetry-java's `integration-tests/osgi`: a new `:osgi-test` module that boots the `semconv` + `semconv-incubating` jars inside a real Apache Felix container and runs JUnit 5 inside it via bnd's `Bundle` → `Resolve` → `TestOSGi` chain. If either bundle's generated OSGi metadata is wrong (bad exports / unsatisfiable imports) the bundle won't resolve and the tests won't run — and the task fails explicitly in that case, since bnd otherwise reports success on zero tests.

Two things surfaced while getting it to resolve, both scoped to the test module:

- The semconv bundles `Import-Package: io.opentelemetry.api.common`, so resolution needs an `opentelemetry-api` that is itself an OSGi bundle. Since OSGi metadata was only added to core in open-telemetry/opentelemetry-java#8417 (well after the `1.33.0` compileOnly baseline this repo pins), the test forces a recent `opentelemetry-api` on its own classpath.
- `opentelemetry-common` (pulled in transitively) declares `Require-Capability: osgi.serviceloader.processor`, so the container needs Aries SPI Fly to resolve even though semconv has no ServiceLoader providers of its own.

Both are documented in comments in `osgi-test/build.gradle.kts`. It's wired into `test`/`check`, so it runs as part of the normal `./gradlew build`.

buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
```diff
@@ -85,6 +95,30 @@ tasks {
...
+          bnd(
+            mapOf(
+              "-exportcontents" to "io.opentelemetry.*",
+              "Import-Package" to importPackages,
+            ),
+          )
+        }
+      }
+    }
+  }
```
trask: In the sibling instrumentation PR (open-telemetry/opentelemetry-java-instrumentation#18995) you moved OSGi into its own `otel.osgi-conventions` plugin, so applying the plugin *is* the opt-in and there's no `osgiEnabled` flag at all. That PR also dropped the whole-project `afterEvaluate` in favour of configuring the task directly. This branch still has the earlier shape.

Here you don't even need a separate plugin — both published modules want OSGi, and `osgi-test` already disables its `jar`. Suggest inlining the bundle config and dropping both properties from `OtelJavaExtension` plus `otelJava.osgiEnabled.set(false)`. (As written the flag doesn't match its doc: with it off, bnd still emits `Bundle-SymbolicName` / `Import-Package` / `Require-Capability` and just turns the exports into `Private-Package`.)

```suggestion
  named<Jar>("jar") {
    // Configure OSGi metadata; BND auto-detects the imports.
    bundle {
      bnd(mapOf("-exportcontents" to "io.opentelemetry.*"))
    }
  }
```

royteeuwen: Good catch on the flag — you're right that with bnd applied, `osgiEnabled=false` didn't actually suppress the bundle manifest, it just demoted the exports to `Private-Package`, so the doc was misleading. Done as suggested: inlined the `jar` bundle config (`-exportcontents: io.opentelemetry.*`, BND auto-detecting imports), dropped the whole-project `afterEvaluate`, and removed both `osgiEnabled` and `osgiOptionalPackages` from `OtelJavaExtension` plus the `otelJava.osgiEnabled.set(false)` in `osgi-test` (its `jar` is disabled anyway). Verified the generated manifest is unchanged — `opentelemetry-semconv` still exports `io.opentelemetry.semconv` and imports `io.opentelemetry.api.common` — and the OSGi resolve test still passes.

buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
```diff
@@ -85,6 +93,13 @@ tasks {
     }
   }
 
+  named<Jar>("jar") {
+    // Configure OSGi metadata; BND auto-detects the imports.
+    bundle {
+      bnd(mapOf("-exportcontents" to "io.opentelemetry.*"))
```
trask: @royteeuwen I'm not sure about this AI suggestion, so just passing it along...


These jars import `io.opentelemetry.api.common` with no version range, so OSGi will wire them to any version of it — including a future 2.x. From the built manifest, next to an import that *does* get a range:

```
Import-Package: io.opentelemetry.api.common,io.opentelemetry.semconv;version="[1.43,2)",...
```

semconv gets one because it's a bnd bundle in this build; the pinned `opentelemetry-api` 1.33.0 isn't, so bnd has no version to read. Setting the range explicitly fixes it — `1.33` is the baseline `:dependencyManagement` already pins. Verified both jars come out with `version="[1.33,2)"` and `:osgi-test:test` still passes.

```suggestion
      bnd(
        mapOf(
          "-exportcontents" to "io.opentelemetry.*",
          "Import-Package" to "io.opentelemetry.api.*;version=\"[1.33,2)\",*"
        )
      )
```

settings.gradle.kts
```diff
@@ -60,3 +60,4 @@ rootProject.name = "semantic-conventions-java"
 include(":dependencyManagement")
 include(":semconv-incubating")
 include(":semconv")
+include(":osgi-test")
```
trask, line 63: The description is out of sync with what landed. It still lists `osgiEnabled` / `osgiOptionalPackages` on `OtelJavaExtension`, but the last commit removed those and `OtelJavaExtension.kt` isn't in the diff. It also doesn't mention this new `:osgi-test` module or the Felix test — Testing just says `./gradlew build` passes. Could you refresh those sections?

trask (approved):

buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
royteeuwen: Good call, applied. Confirmed both jars now import `io.opentelemetry.api.common;version="[1.33,2)"` (and incubating keeps its `io.opentelemetry.semconv;version="[1.43,2)"`), and `:osgi-test` still resolves/passes. Left `*` as the fallback so java.* etc. stay auto-detected.

settings.gradle.kts
royteeuwen: Refreshed the description: dropped the `osgiEnabled`/`osgiOptionalPackages` mention (config is applied unconditionally now), documented the `[1.33,2)` import range, added the `:osgi-test` module + Felix test under Changes/Testing, and updated the manifest excerpts to the current version/range.

